Issue #21, opened 2024-12-04
How to clean the cache


Mic92:
Does this service also support cleaning caches? I only see code that adds more content.

kalbasit:
Great point @Mic92, it's actually on the roadmap (I have not posted the roadmap). I'm thinking of adding a boltdb-like file to the cache to keep track of when cache paths were last pushed and last used to allow for garbage collection; The garbage collector will run on another port so I can use it from a Kubernetes CronJob without having to make it visible to the outside. What do you think about doing it this way?

Mic92:
Sounds good, would the port access the same boltdb? You probably also can just do a simple LRU cache since for a passthru cache references don't matter as much.

kalbasit:
@Mic92 the boltdb idea is to implement LRU, in a way. Of course,  I could make it automatically remove whatever falls outside the allowed size or do it via GC.

For now, I added `DELETE` routes, albeit un-authenticated.

kalbasit:
@Mic92 this is now implemented, do you mind checking? I ended up going with a simple cron-based implementation instead of another server to listen to I figured not everyone runs k8s and this would be a lot simpler to adopt than to have to run a cron in another way.

Mic92:
Nice. Maybe you want an index on `hash` eventually as it's used in every `touchNarInfoQuery`/`touchNarQuery`. @brianmcgee also recently showed me sqlc, [example](https://github.com/Mic92/niks3/blob/main/pg/query.sql), which saves quite a bit of boilercode to write. LRU caching on nar/narinfo for a passthru cache is fine, but might be a bit rough, if you actually plan to write to a binary cache yourself, as it will not take any dependencies into account. In [niks3](https://github.com/Mic92/niks3), we have the concept of [closures](https://github.com/Mic92/niks3/blob/c8b20021394bf2fa273d9f90e8f8242af59962a4/pg/migrations/20241026095416_initial_model.sql#L23). In our case a closure will be the high-level store path that is pushed to a binary cache, the closure will contain the store path itself and all its dependence. Multiple closures can refer to the same objects. But I am also fine testing it as a pure pull through cache. My hope is that we have a solution that we can provide to providers so that cache.nixos.org can save some traffic.

Mic92:
One more question. Do you parallelize querying upstream caches?

kalbasit:
[...]

PR #23, opened 2024-12-04
pkg/server: Respond to  DELETE for  nar and narinfo files

kalbasit:
Add a new API to respond to:
- `DELETE /<hash>.narinfo`
- `DELETE /<hash>.nar[.<compression>]`

This API is disabled by default. It can be enabled with the `--allow-delete` flag.

ref #21

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -163,6 +163,11 @@ func (c *Cache) PutNar(_ context.Context, hash, compression string, r io.ReadClo
 	return err
 }
 
+// DeleteNar deletes the nar from the store.
+func (c *Cache) DeleteNar(_ context.Context, hash string, r io.ReadCloser) error {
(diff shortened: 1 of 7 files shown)


pkg/cache/cache.go
```diff
@@ -163,6 +163,11 @@ func (c *Cache) PutNar(_ context.Context, hash, compression string, r io.ReadClo
 	return err
 }
 
+// DeleteNar deletes the nar from the store.
+func (c *Cache) DeleteNar(_ context.Context, hash, compression string) error {
```
Mic92, line 167: How is that authenticated?

kalbasit: It's not! I am however adding a flag to the server (in this pr, working on it) `--allow-delete` for MVP version. We can add basic auth or some other auth later. I'm not planning to use it, but it's here if needed.

PR #77, opened 2024-12-12
devshells: Add Delve for debugging

kalbasit:
Add Delve to the devshells; Disable nix hardening for fortify otherwize delve does not start. See: https://wiki.nixos.org/wiki/Go#Using_cgo_on_NixOS


kalbasit:
Delve is not working. Relevant discussion: https://discourse.nixos.org/t/golang-delve-debugger-with-nix-flake-issue/22740

kalbasit:
I don't think that this is possible with Devbox so I'm probably going to have to adopt flake early on.

PR #193, opened 2025-03-07
feat(cache): support custom priority

numinit:
Loading the priority from the nix-cache-info at service start results in a service failure if we are unable to load the priority. Allow a custom priority in the URL, similar to how Nix works.

### DIFF
--- a/pkg/cache/upstream/cache.go
+++ b/pkg/cache/upstream/cache.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strconv"
 	"time"
 
 	"github.com/nix-community/go-nix/pkg/narinfo"
@@ -87,9 +88,24 @@ func New(ctx context.Context, u *url.URL, pubKeys []string) (Cache, error) {
 		c.publicKeys = append(c.publicKeys, pk)
 	}
 
-	priority, err := c.parsePriority(ctx)
-	if err != nil {
-		return c, fmt.Errorf("error parsing the priority for %q: %w", u, err)
+	var priority uint64
+
+	if u.Query().Has("priority") {
+		parsedPriority, err := strconv.ParseUint(u.Query().Get("priority"), 10, 16)
+		if err != nil {
+			return c, fmt.Errorf("error parsing the priority from the URL %q: %w", u, err)
+		}
+
+		priority = parsedPriority
+	}
+
+	if priority < 1 {
+		parsedPriority, err := c.parsePriority(ctx)
+		if err != nil {
+			return c, fmt.Errorf("error parsing the priority for %q: %w", u, err)
+		}
+
+		priority = parsedPriority
 	}
 
 	c.priority = priority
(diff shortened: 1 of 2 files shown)


cmd/serve.go
```diff
@@ -185,6 +186,14 @@ func getUpstreamCaches(ctx context.Context, cmd *cli.Command) ([]upstream.Cache,
 			return nil, fmt.Errorf("error parsing --upstream-cache=%q: %w", us, err)
 		}
 
+		var priority int64 = -1
```
kalbasit: Is zero a valid priority? I feel like 0 is invalid and thus you should just start with zero and send uint64 everywhere. If not then fair enough.

pkg/cache/upstream/cache.go
```diff
@@ -52,7 +52,7 @@ type Cache struct {
 	publicKeys []signature.PublicKey
 }
 
-func New(ctx context.Context, u *url.URL, pubKeys []string) (Cache, error) {
+func New(ctx context.Context, u *url.URL, pubKeys []string, priority int64) (Cache, error) {
```
kalbasit: Why not have the parsePriority passes it from the url instead of passing it from main?

kalbasit (changes requested):

pkg/cache/upstream/cache.go
numinit: Ah, very good point - much cleaner!

kalbasit (approved):

numinit:
Thanks for the merge!

Out of curiosity, what happens if two caches have the same priority? Are they currently tried in parallel and the first one to the finish wins, or is it in serial? (This may determine what I contribute next...)

BTW, you'll enjoy the Planet Nix talk where I successfully used ncps to fetch derivations over a mesh network :-)

kalbasit:
> Thanks for the merge!
> 
> Out of curiosity, what happens if two caches have the same priority? Are they currently tried in parallel and the first one to the finish wins, or is it in serial? (This may determine what I contribute next...)

The behavior for selecting an upstream cache is in [this function](https://github.com/kalbasit/ncps/blob/1a8b6a9de0be32dd49bd200a754245e1406b43d2/pkg/cache/cache.go#L1469). My first iteration was to respect the priority and try them one by one sequentially, but this proved to be quite slow. So now all upstreams are tried each in a goroutine and whichever returns 200 first wins; Obviously this isn't great because an upstream with lower priority could be selected and this algorithm could be improved to honor priority but still reaches out to them in parallel.

> BTW, you'll enjoy the Planet Nix talk where I successfully used ncps to fetch derivations over a mesh network :-)

Oh awesome, do you have a link to the talk; I'm interested in watching it!

numinit:
I guess I've got my next contribution in mind :-) 

They're still cutting them together - I see them going up on https://www.youtube.com/channel/UCN2nbMPLJWv3Y__4JuF_hMQ but Adventures in Mesh Networking with NixOS isn't up yet.

PR #310, opened 2025-09-22
feat(auth): Add netrc support for upstream authentication

### LINKED ISSUE #307: ncps netrc support
Hi there

Thanks for creating this :)

I use a self-hosted private external attic instance that machines use with a netrc-file for credentials.

When adding this cache to my ncps local cache it does not see the attic cache and errors out.

Is the netrc-file being used in this instance or is this not supported?

Thanks

_Originally posted by @mannp in https://github.com/kalbasit/ncps/discussions/223_

aciceri:
@kalbasit this is a proof of concept (the netrc path is still hardcoded) but seems to work. What do you think?

It works even if the netrc entry is without the `username`, exactly as nix does.

Since I'm not a go developer I got help form an LLM but I've double checked everything and it makes sense to me.

Assuming that you like what I did: how do you think that we should retrieve the netrc path? I see the following solutions:
- we add an extra CLI argument for the netrc path
- we call `nix show-config --json` (or do we parse `/etc/nix/nix.conf`?) and use the netrc file specified there

Closes #307

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -8,11 +8,13 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
 	"regexp"
 	"time"
 
 	"github.com/robfig/cron/v3"
 	"github.com/rs/zerolog"
+	"github.com/sysbot/go-netrc"
 	"github.com/urfave/cli/v3"
 	"golang.org/x/sync/errgroup"
 
@@ -27,6 +29,32 @@ import (
 
 // ErrCacheMaxSizeRequired is returned if --cache-lru-schedule was given but not --cache-max-size.
 var ErrCacheMaxSizeRequired = errors.New("--cache-max-size is required when --cache-lru-schedule is specified")
+
+// getDefaultNetrcPath returns the default path to the netrc file.
+func getDefaultNetrcPath() string {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return ".netrc" // fallback to current directory
+	}
+
+	return filepath.Join(homeDir, ".netrc")
+}
+
+// parseNetrcFile parses the netrc file and returns the parsed netrc object.
+func parseNetrcFile(netrcPath string) (*netrc.Netrc, error) {
+	file, err := os.Open(netrcPath)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	n, err := netrc.Parse(file)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing netrc file: %w", err)
+	}
+
+	return n, nil
+}
 
 func serveCommand() *cli.Command {
 	return &cli.Command{
@@ -125,6 +153,12 @@ func serveCommand() *cli.Command {
 				Usage:   "Set to host:public-key for each upstream cache",
 				Sources: cli.EnvVars("UPSTREAM_PUBLIC_KEYS"),
 			},
+			&cli.StringFlag{
+				Name:    "netrc-file",
+				Usage:   "Path to netrc file for upstream authentication",
+				Sources: cli.EnvVars("NETRC_FILE"),
+				Value:   getDefaultNetrcPath(),
+			},
 		},
 	}
 }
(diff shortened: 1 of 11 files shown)


kalbasit:
nice @aciceri it looks good!

We should provide a command line flag `--netrc-file` that defaults to `$HOME/.netrc`. You can get to the use's homedir with [this function](https://pkg.go.dev/os#UserHomeDir). In my opinion we should not add support for parsing netrc from nix as not everyone runs it in a Nix environment; For instance, I run ncps on my Kubernetes cluster in a container.

aciceri:
Oh right, I didn't think about it. I'll add a `--netrc-file` argument that defaults to `$HOME/.netrc` then. If the file doesn't exist, both if you explicitly set the argument or leave the default, I'll just print out a warning, ok? 

After having done this I just update the README and the PR is ready? I saw that the NixOS module lives in nixpkgs (I may open a PR there too). Do we want to add a test?

However the CI is failing for some reasons not due to my commit, perhaps missing secrets on external forks.

kalbasit:
> I'll just print out a warning, ok?

It's not really necessary. None of my upstream caches needs a `.netrc` so warning would not help but confuse someone like me. We could however, if we get a 401 mention the netrc feature.

> After having done this I just update the README and the PR is ready?

That's correct, at that point make it ready for review and I will test it/review it before I merge it.

> I saw that the NixOS module lives in nixpkgs (I may open a PR there too). Do we want to add a test?

That's also correct but know that I won't merge the PR upstream until I release a new version of ncps, so make it a draft pr upstream and tag me. You don't need to add a test for this case as it's going to probably be difficult to do so.

> However the CI is failing for some reasons not due to my commit, perhaps missing secrets on external forks.

Yes that's correct. You're the first PR to ncps from a fork and it seems to be failing. I will use your PR to fix it :)

Thanks again for working on this @aciceri

aciceri:
At the end I came up with this message:
```
2025-09-23T22:51:42+02:00 WRN failed to parse netrc file, proceeding without netrc authentication error="open netr: no such file or directory" cmd=serve
```

To me it's clear enough, honestly I would prefer knowing if the netrc is being used or not, it would help me a lot if for some reason I see that one of my authenticated cached doesn't seem to work.

Otherwise the message is something like:
```
2025-09-23T22:53:20+02:00 INF loaded netrc authentication credentials cmd=serve hostname=mycache.tld username=
```

The netrc file is parsed only once and then the credentials are injected into the various cache objects, I hope it's good as approach.
I've also added a mention to the new CLI argument in the README.

cmd/serve.go
```diff
@@ -28,6 +30,32 @@ import (
 // ErrCacheMaxSizeRequired is returned if --cache-lru-schedule was given but not --cache-max-size.
 var ErrCacheMaxSizeRequired = errors.New("--cache-max-size is required when --cache-lru-schedule is specified")
 
+// getDefaultNetrcPath returns the default path to the netrc file.
+func getDefaultNetrcPath() string {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return ".netrc" // fallback to current directory
```
kalbasit: Hmm I'm not sure I agree to this. I don't see anyone setting their PWD to where `.netrc` is before running ncps (as a daemon) but I do see someone relying on ncps finding it in the user's home (if `ncps serve --help` told them so). I would panic here instead of returning a string.

pkg/cache/upstream/cache_test.go
kalbasit, line 1: It would be nice to add a test for this functionality. You can easily create an httptest server and protect it with basic auth, here's a basic example you can use for inspiration: https://go.dev/play/p/lW0I1fMU0iV

kalbasit:
@aciceri also no need to maintain a good commit history or force push since I will squash commit anyways so the commit history (although preserved in this PR) is lost on the main branch.

kalbasit:
@aciceri alright I fixed the tests in #316 and I merged main into your pr to test it.

Although, fyi you have a legitimate failing test to fix, just a missed New() change

```
ncps> # github.com/kalbasit/ncps/pkg/cache/healthcheck_test [github.com/kalbasit/ncps/pkg/cache/healthcheck.test]
ncps> pkg/cache/healthcheck/healthcheck_test.go:34:76: not enough arguments in call to upstream.New
ncps>   have (context.Context, *url.URL, []string)
ncps>   want (context.Context, *url.URL, []string, *upstream.NetrcCredentials)
```

cmd/serve.go
aciceri: My bad, in my head that code was doing a different thing...

aciceri:
As soon as you'll merge #315 I'll rebase onto `main` and I'll be able to build the checks locally. However in theory it should pass now.

kalbasit:
Thank you, @aciceri this looks amazing! I pushed a nitpick to keep the flags ordered alphabetically but otherwise it's ready to merge. I will let GH do the honors.

aciceri:
Thank you for this great piece of software!
I'll let you chase all the missing tests at this point, meanwhile I'll try to adjust the NixOS module and actually use this version on my system, indeed for now I've only tried it playing with `nix copy` and `attic push`.

PR #539, opened 2026-01-13
fix: prevent double-counting of shared NAR files during LRU cleanup

### LINKED ISSUE #470: LRU will not clean-up as expected now that narinfo n-n nar_files
high

The `GetLeastUsedNarInfos` query, used to select items for LRU eviction, appears to have a flaw in its size calculation. It sums the `file_size` for each `narinfo` reference, which will lead to double-counting the size of shared `nar_file`s. This can cause the LRU logic to evict more `narinfo` records than necessary while failing to free up the expected amount of disk space, as a shared `nar_file` will not be deleted if it's still referenced by other non-evicted `narinfo`s.

For example, if `narinfo1` and `narinfo2` both point to a 1GB `nar_file_A` (which is also referenced by a newer `narinfo3`), the query will count 2GB towards eviction. If the cleanup target is 1.5GB, both `narinfo1` and `narinfo2` might be deleted, but `nar_file_A` will remain, freeing 0 bytes of storage instead of the anticipated amount.

The eviction logic should be revised to accurately account for shared NARs to ensure effective space management.

_Originally posted by @gemini-code-assist[bot] in https://github.com/kalbasit/ncps/pull/464#discussion_r2660507653_

kalbasit:
The GetLeastUsedNarInfos query previously summed the file_size for every
narinfo record pointing to a nar_file. When multiple narinfo records
shared the same nar_file, its size was counted multiple times, leading
to an inflated total size calculation and premature eviction of metadata
without freeing the expected disk space.

This implementation refactors the query to use an IN clause with a
subquery that identifies the unique set of nar_file IDs associated with
the candidate narinfo records. This ensures that the total size
calculation accurately reflects the physical disk space utilized.

The fix is verified with a new regression test TestRunLRUWithSharedNar
which simulates a shared NAR file scenario and confirms correct eviction
behavior.

fixes #470


kalbasit:
* **#539** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/539?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/539?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #540, opened 2026-01-13
fix: handle upstream timeouts as 404 in HasNar and HasNarInfo

### LINKED ISSUE #298: upstream.HasNar times out ends up in a 500
I got an http status 500 and when I looked at the trace it seems that upstream.HasNas has 404 for one upstream and timed out for another; The timeout ended up returning a 500; I think a  timeout on HasNar should return a 404 instead.

[Trace-8df479-2025-09-03 22_32_13.json](https://github.com/user-attachments/files/22131671/Trace-8df479-2025-09-03.22_32_13.json)

Image

kalbasit:
A timeout when checking an upstream for a NAR or NarInfo should be treated
as if the resource was not found. This prevents a single slow or responsive
upstream from causing the entire request to fail with an HTTP 500 error.

The implementation catches context.DeadlineExceeded and net.Error timeouts
in the HasNar and HasNarInfo methods of the upstream cache and returns
false (not exists) instead of propagating the error.

fixes #298

### DIFF
--- a/pkg/cache/upstream/cache.go
+++ b/pkg/cache/upstream/cache.go
@@ -366,12 +366,7 @@ func (c *Cache) HasNarInfo(ctx context.Context, hash string) (bool, error) {
 
 	resp, err := c.doRequest(ctx, http.MethodHead, u)
 	if err != nil {
-		if errors.Is(err, context.DeadlineExceeded) {
-			return false, nil
-		}
-
-		var netErr net.Error
-		if errors.As(err, &netErr) && netErr.Timeout() {
+		if isTimeout(err) {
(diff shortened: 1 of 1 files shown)


kalbasit:
* **#540** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/540?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/540?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #543, opened 2026-01-13
fix: deadlock race condition between narinfo/nar jobs

mrene:
Some calls to retrieve a narinfo were hanging our ncps instance. It might be due to multiple ci runners hitting the same file multiple times, but I could even reproduce it with a single curl call (in the cluster though, so other pods might have been hammering it at the same time).

```text
$ curl -vvv 'http://ncps.ci.svc.cluster.local/3wdd686i19ak2hgsdlikn3vm2cfyggip.narinfo'
*   Trying 10.3.253.225:80...
* Connected to ncps.ci.svc.cluster.local (10.3.253.225) port 80 (#0)
> GET /3wdd686i19ak2hgsdlikn3vm2cfyggip.narinfo HTTP/1.1
> Host: ncps.ci.svc.cluster.local
> User-Agent: curl/7.81.0
> Accept: */*
>
<hangs here>
```

What seems to happen (according to claude)
  1. Narinfo request comes in for abc123
  2. coordinateDownload creates upstreamJobs["abc123"] = ds for narinfo
  3. Narinfo goroutine starts, downloads narinfo successfully
  4. Narinfo goroutine then calls prePullNar to fetch the NAR
  5. NAR's coordinateDownload checks upstreamJobs["abc123"] - finds narinfo's entry!
  6. NAR download thinks "someone else is already downloading this" and waits on ds.stored
  7. But ds.stored is for the narinfo, which already completed
  8. Deadlock: NAR waits forever on a channel that will never be signaled for NAR

This uses a different lock key for the nar and narinfo downloads, which solves the problem on our side.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -39,6 +39,12 @@ const (
 	cacheLockKey         = "cache"
 )
 
+// narInfoJobKey returns the key used for tracking narinfo download jobs.
+func narInfoJobKey(hash string) string { return "download:narinfo:" + hash }
+
+// narJobKey returns the key used for tracking NAR download jobs.
+func narJobKey(hash string) string { return "download:nar:" + hash }
+
 var (
 	// ErrHostnameRequired is returned if the given hostName to New is not given.
 	ErrHostnameRequired = errors.New("hostName is required")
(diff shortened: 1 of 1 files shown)


kalbasit:
@mrene Thank you very much for debugging this issue and coming up with a fix for it! I wanted to add a test to your fix but I wasn't able to because you probably did not check `Allow edits from maintainers`. I filed #545 and cherry-picked your commit. I will get it merged and close this one. Thanks again!

mrene:
Thanks for the quick merge!

PR #544, opened 2026-01-14
fix: resolve deadlock between narinfo and nar downloads

kalbasit:
fix: resolve deadlock between narinfo and nar downloads

This change resolves a deadlock race condition that occurred when pullNarInfo triggered a prePullNar operation (e.g., for none compression) with the same hash key.

Why:
When pullNarInfo and prePullNar used the same hash as a key in the upstreamJobs map, they could enter a circular dependency where pullNarInfo holds the job (waiting for prePullNar) while prePullNar waits for the same job to complete.

How:
- Introduced narInfoJobKey and narJobKey helpers in pkg/cache/cache.go to ensure distinct keys in the upstreamJobs map, preventing collisions even if the underlying hashes are identical.
- Created TestDeadlock_NarInfo_Triggers_Nar_Refetch in pkg/cache/cache_deadlock_test.go which reproduces the scenario by forcing a hash collision between a NarInfo and its Nar URL, asserting that the operation completes without timeout.

revert the actual fix

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -39,12 +39,6 @@ const (
 	cacheLockKey         = "cache"
 )
 
-// narInfoJobKey returns the key used for tracking narinfo download jobs.
-func narInfoJobKey(hash string) string { return "download:narinfo:" + hash }
-
-// narJobKey returns the key used for tracking NAR download jobs.
-func narJobKey(hash string) string { return "download:nar:" + hash }
-
 var (
 	// ErrHostnameRequired is returned if the given hostName to New is not given.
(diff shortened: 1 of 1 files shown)


kalbasit:
* **#544** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/544?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/544?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #545, opened 2026-01-14
fix: resolve deadlock between NarInfo and NAR download jobs

kalbasit:
# Goal

This PR resolves a deadlock race condition that occurs when a NarInfo download triggers a NAR download with the same hash key (common when compression is set to none).

## Problem

Both NarInfo and NAR download jobs were using the same raw hash as a key in the upstreamJobs map and for the distributed locker. When pullNarInfo triggered a prePullNar operation for the same hash, they would both attempt to acquire the same lock or wait on the same job entry, leading to a circular dependency and a permanent hang.

## Proposed Changes

- Namespace Job Keys: Introduced narInfoJobKey and narJobKey helper functions in cache.go to provide distinct namespaces for NarInfo and NAR jobs.
- Update Job Tracking: Updated prePullNarInfo and prePullNar to use these namespaced keys when calling coordinateDownload.
- Update Cleanup Logic: Ensured that the cleanup logic in both pullNarInfo and pullNarIntoStore uses the correct namespaced key when removing entries from the upstreamJobs map.

### Automated Tests

- Added a regression test TestDeadlock_NarInfo_Triggers_Nar_Refetch in pkg/cache/cache_deadlock_test.go.
- This test deliberately forces a hash collision between a NarInfo and its associated NAR URL to reproduce the deadlock scenario.
- Verified that the test passes without timing out.

closes #543


kalbasit:
* **#545** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/545?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/545?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #648, opened 2026-01-24
test: fix flaky cache tests and address connection leaks

kalbasit:
This commit fixes a race condition in TestGetNarInfo_MigratesInvalidURL and
addresses a connection leak in TestDeadlock_ContextCancellation_DuringDownload.

In TestGetNarInfo_MigratesInvalidURL, the test was checking for NarInfo
deletion from storage immediately after a background migration was triggered.
Since the deletion happens asynchronously in a goroutine, this caused
occasional failures. Wrapping the check in assert.Eventually ensures the
test waits for the migration to complete.

In TestDeadlock_ContextCancellation_DuringDownload, the test server was
sometimes blocked in Close because the slow handler was still active
after the request context was canceled. Adding a synchronization channel
ensures the handler has finished before the server is closed.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -334,6 +334,9 @@ type Cache struct {
 	upstreamJobs   map[string]*downloadState
 
 	cron *cron.Cron
+
+	// Wait group to track background operations
+	backgroundWG sync.WaitGroup
 }
 
 type downloadState struct {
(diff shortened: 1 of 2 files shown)


kalbasit:
* **#648** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/648?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/648?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #581, opened 2026-01-18
Migrate NarInfo storage to de-normalized database tables


kalbasit:
Currently, ncps stores .narinfo files in external storage (Local Filesystem or S3) while maintaining only basic metadata (hash and timestamps) in the database.

The Problem: Since NarInfo files are typically very small (under 1KB), retrieving them from storage introduces significant overhead:

- Latency: Fetching from S3 adds ~100-200ms per request, compared to ~1-10ms for a database lookup.
- Storage Management: Managing millions of small files can lead to inode exhaustion on local filesystems and high object counts (and associated API costs) in S3.
- Consistency: Maintaining data in two places (Storage + DB) increases the risk of metadata drift.

The Proposal: I propose moving the NarInfo content exclusively into the database using a de-normalized schema. This will make the database the single source of truth for all metadata.

Proposed Changes:

1. Schema Update: Add columns to the narinfos table to store all standard fields: StorePath, URL, Compression, FileHash, FileSize, NarHash, NarSize, Deriver, System, and CA.
2. Helper Tables: Create dedicated tables for multi-value fields to allow for efficient querying:
  a. narinfo_references: To store and query store path dependencies.
  b. narinfo_signatures: To store and audit signatures.
3. Cache Refactor: Update ncps to fetch/store NarInfo data directly in these tables, bypassing the NarInfoStore.
4. Migration CLI: Introduce a command to migrate existing .narinfo files from storage into the database.

Benefits:

- Performance: Significant reduction in NarInfo retrieval time, especially for S3 backends.
- Queryability: Enables advanced database-native queries (e.g., "Which packages reference this hash?") without parsing text files.
- Simplicity: Reduces the complexity of storage management and synchronization.

Summary of Fields to be De-normalized:
Based on discovery, the following fields will be mapped:

- CA, Compression, Deriver, FileHash, FileSize, NarHash, NarSize, StorePath, URL.
- References (via helper table).
- Sig (via helper table).

PR #652, opened 2026-01-26
fix: resolve race condition in GetNarInfo during migration

kalbasit:
This fixes a TOCTOU (Time-of-check-time-of-use) race condition that occurred
when GetNarInfo was called while a concurrent migration was deleting narinfo
files from storage.

The race scenario:
1. GetNarInfo checks database -> not found (NULL URL from placeholder)
2. GetNarInfo checks HasNarInfo -> true (file exists in storage)
3. Migration runs concurrently: writes full data to DB, deletes from storage
4. GetNarInfo tries to read from storage -> fails (file already deleted)
5. Without fix: error returned to user, even though migration succeeded

The fix implements a retry mechanism: when storage fetch fails with ErrNotFound,
GetNarInfo retries the database lookup to catch successfully completed migrations.
This handles the race gracefully without blocking concurrent operations or adding
additional locks.

Changes:
- Added TestGetNarInfo_RaceConditionDuringMigrationDeletion test that precisely
  reproduces the race using a hook wrapper
- Added handleStorageFetchError helper function to implement retry logic
- Modified GetNarInfo to call the new helper when storage fetch fails
- Reduced nesting complexity to satisfy nestif linter

All existing tests pass, confirming the fix doesn't break existing functionality.

Part of #581

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -1385,13 +1385,7 @@ func (c *Cache) GetNarInfo(ctx context.Context, hash string) (*narinfo.NarInfo, 
 
 			// If narinfo was purged, continue to fetch from upstream
 			if !errors.Is(err, errNarInfoPurged) {
-				retryErr := c.handleStorageFetchError(ctx, hash, err, &narInfo, &metricAttrs)
-				if retryErr != nil {
-					return retryErr
-				}
-
-				// Retry succeeded, return success
-				return nil
+				return c.handleStorageFetchError(ctx, hash, err, &narInfo, &metricAttrs)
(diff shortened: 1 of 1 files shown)


kalbasit:
* **#652** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/652?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/652?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #658, opened 2026-01-26
fix: cleanup legacy narinfo in PutNarInfo to avoid race condition

kalbasit:
In a thundering herd scenario where multiple concurrent GetNarInfo and PutNarInfo calls occur for the same hash, PutNarInfo might commit to the database before GetNarInfo triggers a migration. This skips the migration logic, leaving legacy narinfo files in storage, which causes tests to fail and leads to storage inconsistency.

This change adds an explicit check for legacy narinfo files in PutNarInfo. If a file exists in the storage backend after the database update, it is now deleted. This ensures consistency regardless of which operation finishes first.

I have also added a deterministic test  that uses hooks to force this race condition and verify that the cleanup always occurs.

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -2429,7 +2429,7 @@ func TestGetNarInfo_RaceWithPutNarInfoDeterministic(t *testing.T) { //nolint:par
 	require.NoError(t, err)
 
 	// Ensure PutNarInfo actually ran
-select {
+	select {
 	case <-putFinished:
 		// ok
 	case <-time.After(5 * time.Second):


kalbasit:
* **#658** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/658?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/658?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #733, opened 2026-02-04
perf: optimize NAR HEAD requests by using NarInfo FileSize

kalbasit:
This optimization leverages the FileSize from the database (via NarInfo)
to fulfill HEAD requests for NAR files. By providing the size directly
from the cache, we avoid unnecessary fetching of the NAR file from
storage, significantly improving response times for HEAD requests.

Additionally, this change fixes a bug in checkAndFixNarInfo where it
would prematurely pull NARs from upstream to verify their size, even if
the NAR was not already in the local cache.

Changes:
- Added Cache.GetNarFileSize to retrieve the file size from the database.
- Updated Server.getNar to use GetNarFileSize for HEAD requests.
- Optimized HEAD response to return 200 OK with Content-Length when the size
  is known, instead of StatusNoContent which could strip the header.
- Fixed checkAndFixNarInfo to only verify NAR size if it already exists in
  one of the local stores.
- Integrated TestGetNar_HeadOptimization into server_test.go.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -339,7 +339,7 @@ func (s *Server) getNarInfo(withBody bool) http.HandlerFunc {
 		h.Set(contentLength, strconv.Itoa(len(narInfoBytes)))
 
 		if !withBody {
-			w.WriteHeader(http.StatusNoContent)
+			w.WriteHeader(http.StatusOK)
 
 			return
 		}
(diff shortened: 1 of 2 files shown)


kalbasit:
* **#733** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/733?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/733?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#735** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/735?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #757, opened 2026-02-05
chore: update flake.lock and run go mod tidy

kalbasit:


kalbasit:
@gemini-code-assist wow you got this one so wrong. This is just a `go mod tidy` that cleaned-up the go.sum file without updating dependencies or anything else in the code base...

PR #762, opened 2026-02-05
fix: avoid MySQL deadlocks in migrate-nar-to-chunks tests

kalbasit:
The test was failing with MySQL deadlock errors (Error 1213) when
migrating multiple narinfos concurrently. This is caused by concurrent
INSERT ... ON DUPLICATE KEY UPDATE operations on the nar_files table,
which can trigger deadlocks in InnoDB even when inserting different rows.

The fix sets --concurrency=1 for the migrate-narinfo command in tests
that process multiple NARs, forcing sequential processing and preventing
the race condition that causes deadlocks. This approach is similar to
the fix in commit 17d41233 for the CDC flaky test.

The deadlock only occurred with MySQL; SQLite and PostgreSQL tests were
unaffected.

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/pkg/ncps/migrate_nar_to_chunks_test.go
+++ b/pkg/ncps/migrate_nar_to_chunks_test.go
@@ -177,10 +177,12 @@ func testMigrateNarToChunksDryRun(factory narToChunksMigrationFactory) func(*tes
 		require.NoError(t, err)
 
 		// Migrate NarInfo to DB first
+		// Use --concurrency=1 to avoid MySQL deadlocks when migrating multiple narinfos in parallel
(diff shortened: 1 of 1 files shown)


kalbasit:
* **#762** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/762?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/762?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #774, opened 2026-02-06
fix: GetNar timeout when CDC is disabled

kalbasit:
Fix a 30s timeout in TestCacheBackends/MySQL/GetNar where GetNar would mistakenly enter the progressive chunk streaming path if a database record existed, even if CDC was disabled.

The issue occurred because GetNar and prePullNar only checked for the existence of a nar_files record without verifying if CDC was enabled. A preceding GetNarInfo call would create this record, causing subsequent GetNar calls to wait for chunks that would never be produced.

This PR:
- Adds isCDCEnabled(), getChunkStore(), and getCDCInfo() helper methods to safely access CDC-related fields with proper locking.
- Ensures GetNar and prePullNar only use chunk-related logic when isCDCEnabled() returns true.
- Adds proper locking to SetChunkStore to prevent race conditions.
- Refactors all manual CDC configuration checks to use the new helper methods for better consistency and thread safety.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -526,6 +526,9 @@ func (c *Cache) SetCDCConfiguration(enabled bool, minSize, avgSize, maxSize uint
 
 // SetChunkStore sets the chunk store.
 func (c *Cache) SetChunkStore(cs chunk.Store) {
+	c.cdcMu.Lock()
+	defer c.cdcMu.Unlock()
+
 	c.chunkStore = cs
 }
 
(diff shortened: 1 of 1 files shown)


kalbasit:
* **#774** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/774?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/774?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #781, opened 2026-02-06
fix: resolve goroutine leak in selectUpstream

kalbasit:
The selectUpstream function used unbuffered channels for upstream results
and errors. When multiple upstreams responded concurrently, goroutines that
tried to send after the first result was consumed would block forever on
the unbuffered channel sends, leaking goroutines permanently.

Fixed by making both channels buffered with len(ucs) capacity so sends
never block and all goroutines can complete cleanly. Added tests that
verify all worker goroutines complete in both the all-succeed and
mixed-success/error scenarios.

### DIFF
--- a/pkg/cache/select_upstream_leak_internal_test.go
+++ b/pkg/cache/select_upstream_leak_internal_test.go
@@ -47,7 +47,7 @@ func createDummyUpstreams(t *testing.T, n int) []*upstream.Cache {
 func TestSelectUpstream_NoGoroutineLeak(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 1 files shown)


kalbasit:
* **#781** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/781?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/781?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #785, opened 2026-02-06
fix: fix bulk operations with multiple slices in gen-db-wrappers

kalbasit:
The bulk operation generator was failing when multiple slice parameters were present because it was using a range loop with a blank identifier for the index, which only allowed accessing the first slice's elements via the loop variable.

This change switches the loop to an indexed range loop (for i := range ...) which allows accessing elements from all slices using the same index. Additionally, it fixes a bug where scalar fields were incorrectly being mapped to the loop variable if their types happened to match, ensuring they are instead correctly mapped from the input argument.

### DIFF
--- a/nix/gen-db-wrappers/src/main.go
+++ b/nix/gen-db-wrappers/src/main.go
@@ -631,27 +631,37 @@ func (w *{{$.Engine.Name}}Wrapper) {{.Name}}({{joinParamsSignature .Params}}) ({
 	{{- end -}}
 
 	{{- if $isAutoLoop -}}
+		{{- $bulkParamType := (index .Params 1).Type -}}
+		{{- $bulkStructInfo := getStruct $bulkParamType -}}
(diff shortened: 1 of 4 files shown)


kalbasit:
* **#786** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/786?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#784** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/784?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#785** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/785?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/785?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #800, opened 2026-02-08
perf: avoid double-reading data during CDC

kalbasit:
The content-defined chunking (CDC) process was previously reading data from disk even though the data was already available in memory within the chunker. This change optimizes the process by passing the chunk data directly from the chunker to the cache layer using a sync.Pool to minimize allocations and ensure thread-safety.

How:
- Added Data field to chunker.Chunk.
- Implemented per-chunker sync.Pool for byte buffers in CDCChunker.
- Updated CDCChunker.Chunk to copy data to a pooled buffer before sending it on the channel.
- Updated storeNarWithCDC in pkg/cache to use the passed data instead of reading from file.
- Added Free() mechanism to return buffers to the pool.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -1313,19 +1313,15 @@ func (c *Cache) storeNarWithCDC(ctx context.Context, tempPath string, narURL *na
 				return totalSize, nil
 			}
 
-			// Process chunk
-			data := make([]byte, chunkMetadata.Size)
-
-			_, err := f.ReadAt(data, chunkMetadata.Offset)
-			if err != nil {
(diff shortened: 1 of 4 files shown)


kalbasit:
* **#800** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/800?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/800?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #892, opened 2026-02-14
feat: store and serve uncompressed NARs as zstd

kalbasit:
Normalize NARs that are originally uncompressed (Compression: none) by storing them using ZSTD compression. This optimization reduces storage requirements while maintaining full compatibility with clients that expect uncompressed data by transparently decompressing upon retrieval.

Key implementation details:
- Storage normalization: storeNarFromTempFile now detects uncompressed NARs and uses an io.Pipe paired with a ZSTD encoder to compress the stream before it hits the store.
- Metadata normalization: pullNarInfo and PutNarInfo ensure that for uncompressed NARs, the FileSize and FileHash in the metadata correctly reflect the uncompressed NarSize and NarHash, preventing serialization inconsistencies.
- Transparent retrieval: Updated the loading logic (GetNar, getNarFromStore, etc.) to prioritize checking for .nar.zst files and performing on-the-fly decompression when Compression: none is indicated in the database.
- Test robustness: Updated test suites to handle asynchronous storage via waitForFile and added validation for correct size/hash calculations of normalized entries.

PR #896, opened 2026-02-14
fix: respect Content-Encoding header for NAR compression detection

kalbasit:
Previously, ncps would incorrectly assume that all NARs fetched from upstream
with zstd support enabled would be zstd-compressed, even when the upstream
returned uncompressed data. This caused NARs to be stored with the wrong
compression type in the database.

The fix moves the compression type detection to occur after receiving the
HTTP response, checking the Content-Encoding header to determine the actual
compression used by the upstream server. This ensures that:

1. If upstream returns Content-Encoding: zstd, we store as zstd
2. If upstream returns no Content-Encoding header, we store as the original
   compression type from the narinfo (e.g., 'none')

Added comprehensive tests to verify both scenarios work correctly.

This PR is an adaptation of the stack #892 re-implemented to just solve the bug
instead of what the #892 stack does.

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/http/httptest"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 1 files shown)


kalbasit:
* **#896** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/896?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/896?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `release-0.8`


<h2></h2>

PR #926, opened 2026-02-17
fix: heal orphan NARs, idempotent CDC migration, and reduce migration churn

kalbasit:
This commit fixes four data integrity and robustness issues in the NAR pull
and CDC migration code paths:

**Issue 1 (High):** storeNarFromTempFile silently returned nil when
narStore.PutNar returned ErrAlreadyExists, leaving orphaned NARs in storage
with no DB record. If a process crashed between narStore.PutNar and
ensureNarFileRecord, subsequent requests would forever skip DB record
creation. Fixed by fetching the stored file size and falling through to
ensureNarFileRecord, mirroring the existing behaviour in PutNar.

**Issue 2 (Medium):** getNarFromStore had a TODO comment acknowledging it
should create a DB record when a NAR exists in storage but not in the DB
(crash-orphan state). Without the record, LRU tracking silently skips the
NAR. Fixed by setting a flag on IsNotFoundError and calling
ensureNarFileRecord after the transaction completes.

**Issue 3 (High):** MigrateNarToChunks was not fully idempotent. If the
process crashed between storeNarWithCDC (which atomically commits chunks and
relinks narinfos to the new nar_file) and the subsequent
UpdateNarInfoCompressionAndURL / DeleteNar steps, a second call would detect
hasChunks==true and immediately return ErrNarAlreadyChunked without fixing
the stale narinfo URL or removing the original whole-file NAR. Fixed by
extracting the cleanup logic into migrateNarToChunksCleanup and calling it
both on the normal path and on the already-chunked early-exit path.

**Issue 4 (Low):** maybeBackgroundMigrateNarToChunks was called on every
GetNarInfo database hit, even for NARs already stored as CDC chunks
(compression=none). For Redis-backed distributed deployments this generates a
spurious TryLock + HasNarInChunks round-trip on every cache hit. Fixed by
skipping the migration trigger when narURL.Compression is already none;
Compression:none NARs that still need migration are still caught by the
GetNar path.

As a housekeeping improvement, storeNarWithCDC and storeNarFromTempFile had
their unused int64 return value removed to satisfy the unparam linter.

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

PR #927, opened 2026-02-17
fix: immediately clean up orphaned chunk files during stale CDC lock recovery

kalbasit:
When findOrCreateNarFileForCDC detects a stale chunking lock (>= 1hr old),
it deletes the nar_file_chunks junction entries for the partial chunking
attempt. Previously, the actual chunk files and their chunks DB records were
left on disk as orphans, to be cleaned up only during the next RunLRU GC
cycle — which could be hours or days away.

This commit adds immediate cleanup via cleanupStaleLockChunks:
- Before deleting the junction entries (inside the transaction), collect the
  chunk records that will become orphaned.
- After the transaction commits, call GetOrphanedChunks and intersect with
  the collected chunk IDs to find chunks that are now truly unreferenced
  (not shared with other completed NAR files).
- For each truly orphaned chunk: delete the DB record (DeleteChunkByID)
  then delete the physical file from chunkStore.

Also adds a code comment in storeNarWithCDC explaining the known limitation
that chunk files written before recordChunkBatch commits cannot be found by
the GC if the process crashes between those two steps (Issue 6).

Two tests are added/updated:
- testCDCStaleLockCleansUpChunkFiles: verifies that chunk FILES in chunkStore
  are immediately deleted during stale lock recovery (not just junction entries).
- testCDCMigrateNarToChunksRecoversFromPartialChunking: extended to also assert
  that the orphaned chunk DB record is deleted immediately (not left for GC).

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -1455,7 +1455,7 @@ func (c *Cache) storeNarWithCDC(ctx context.Context, tempPath string, narURL *na
 	narURL.Compression = nar.CompressionTypeNone
 
 	// 1. Create or get NarFile record
-	narFileID, err := c.findOrCreateNarFileForCDC(ctx, narURL, fileSize)
+	narFileID, staleLockChunks, err := c.findOrCreateNarFileForCDC(ctx, narURL, fileSize)
 	if err != nil {
 		if errors.Is(err, storage.ErrAlreadyExists) {
 			zerolog.Ctx(ctx).Debug().Msg("nar already exists in chunk storage, skipping")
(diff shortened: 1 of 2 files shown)


kalbasit:
* **#927** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/927?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/927?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#926** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/926?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

kalbasit:
### Merge activity

* **Feb 17, 11:35 PM UTC**: @kalbasit merged this pull request with [Graphite](https://app.graphite.com/github/pr/kalbasit/ncps/927).

PR #930, opened 2026-02-18
fix: run.py must start internal instances with direnv exec

kalbasit:
When run.py creates a ncps instance in a tmux split panes, it calls
itself with python3 directly which in turn calls go run; However, this
means that the child self-call does not include the environment needed
by ncps and the wrong Go binary will be used.

By using `direnv exec`, it guarentees that it's running in the same
environment as its parent, presuming it was not modified.


kalbasit:
* **#930** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/930?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/930?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #933, opened 2026-02-18
perf: decompress upstream NAR on-the-fly when CDC client piggybacks on xz download

kalbasit:
When CDC is enabled, prePullNarInfo starts a background download of the NAR in
its original upstream compression format (e.g. xz) for two reasons: faster TTFB
from the upstream (which serves pre-compressed files directly) and better CDC chunk
deduplication (chunks are derived from raw uncompressed bytes, so all clients share
the same chunk set regardless of which compression they requested).

However, when a concurrent GetNar(compression=none) request piggybacked on that
in-progress xz download, the streaming goroutine was sending the raw compressed bytes
verbatim to the client — who expected uncompressed data. This caused data corruption
for any Nix client receiving a CDC-normalized NAR during an active background download.

Fix: the streaming goroutine now checks ds.tempFileCompression (set in
pullNarIntoStore to record the actual compression of bytes written to the temp file).
When that compression differs from narURL.Compression (i.e. the temp file holds
compressed bytes but the client wants uncompressed), the goroutine wraps the temp file
in a fileAvailableReader (which blocks until bytes arrive, enabling true streaming
decompression while the download is still in progress) and then passes it through
nar.DecompressReader to decompress on-the-fly into the response pipe.

Test: testCDCFirstPullCompletesBeforeChunking now uses real xz-compressed test data
(created with ulikunitz/xz) served via a MaybeHandler that overrides Nar2's NAR URL.
It asserts both that the response completes before the slow chunker delay (Bug 1) and
that the body equals the original uncompressed content, not the raw xz bytes (Bug 2).

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -384,13 +384,14 @@ type downloadState struct {
 	cond *sync.Cond
 
 	// Information about the asset being downloaded
-	wg           sync.WaitGroup // Tracks active readers streaming from the temp file
-	cleanupWg    sync.WaitGroup // Tracks download completion to trigger cleanup
-	cdcWg        sync.WaitGroup // Tracks CDC background goroutine; zero by default (non-CDC)
-	closed       bool           // Indicates whether new readers are allowed (protected by mu)
-	assetPath    string
-	bytesWritten int64
-	finalSize    int64
+	wg                  sync.WaitGroup // Tracks active readers streaming from the temp file
+	cleanupWg           sync.WaitGroup // Tracks download completion to trigger cleanup
+	cdcWg               sync.WaitGroup // Tracks CDC background goroutine; zero by default (non-CDC)
+	closed              bool           // Indicates whether new readers are allowed (protected by mu)
+	assetPath           string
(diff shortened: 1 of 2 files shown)


kalbasit:
* **#936** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/936?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#935** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/935?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#933** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/933?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/933?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#932** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/932?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #992, opened 2026-02-28
fix: prevent fsck false positives for chunked nar_files when CDC config is absent

kalbasit:
Two bugs compounded to cause all chunked nar_files to be falsely
reported as 'missing from storage' when the cdc_enabled DB config key
could not be read.

Bug 1 — Silent failure of CDC detection: if GetConfigByKey returned an
error, cdcMode stayed false with no log output. The error is now logged
as a warning, and a data-based fallback is added: if
HasAnyChunkedNarFiles returns true, cdcMode is enabled automatically
with a warning prompting the operator to verify --cache-database-url.

Bug 2 — Unnecessary cdcMode gate on the TotalChunks check: the condition
'if cdcMode && nf.TotalChunks > 0' meant that when Bug 1 fired
(cdcMode=false), all 60,953 chunked nar_files fell through to the HasNar
check and were all flagged as missing. Chunked nar_files (total_chunks >
0) are always stored in chunk storage, never as whole NAR files, so the
cdcMode guard is incorrect. The fix removes it: 'if nf.TotalChunks > 0'
skips the whole-NAR check unconditionally.

A new HasAnyChunkedNarFiles DB query is added to all three engine query
files (SQLite, PostgreSQL, MySQL) and regenerated via sqlc + go
generate.

A regression test
(CDC/ChunkedNarFilesNotFlaggedAsMissingWithoutCDCConfig) verifies that
chunked nar_files with no cdc_enabled config key produce zero fsck
issues.

### DIFF
--- a/db/query.mysql.sql
+++ b/db/query.mysql.sql
@@ -366,4 +366,4 @@ FROM chunks;
 
 -- name: HasAnyChunkedNarFiles :one
 -- Returns true if any nar_file has total_chunks > 0 (used for CDC auto-detection).
-SELECT EXISTS(SELECT 1 FROM nar_files WHERE total_chunks > 0);
+SELECT EXISTS(SELECT 1 FROM nar_files WHERE total_chunks > 0) AS "exists";
--- a/db/query.postgres.sql
+++ b/db/query.postgres.sql
@@ -398,4 +398,4 @@ FROM chunks;
 
 -- name: HasAnyChunkedNarFiles :one
 -- Returns true if any nar_file has total_chunks > 0 (used for CDC auto-detection).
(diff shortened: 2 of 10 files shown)


kalbasit:
* **#992** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/992?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/992?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kalbasit:
Successfully created backport PR for `release-0.9`:
- #995

PR #1012, opened 2026-03-02
fix: set deployment strategy to Recreate for SQLite

### LINKED ISSUE #1010: helm: chart should set `.spec.strategy.type` `Recreate` when using `mode: deployment` with sqlite.
Otherwise changes to the deployment spec will block the rollout when the new pods is scheduled on a different node **and** the PVC is RWO.

kalbasit:
The Helm chart now uses the Recreate deployment strategy when mode is
set to deployment and SQLite is being used as the database.

Without this strategy, Kubernetes defaults to RollingUpdate, which can
cause rollouts to hang when a new pod is scheduled on a different node
than the old one but cannot mount the PersistentVolume (PV) because it
is ReadWriteOnce (RWO) and still attached to the old pod. This is
particularly relevant for SQLite which relies on a single PVC for
storage.

The change handles this by conditionally setting strategy.type based on
the database type in deployment.yaml. Unit tests have been added to
verify the correct rendering for both SQLite (Recreate) and other
databases (RollingUpdate).

fixes #1010

### DIFF
--- a/charts/ncps/tests/deployment_test.yaml
+++ b/charts/ncps/tests/deployment_test.yaml
@@ -492,3 +492,14 @@ tests:
       - equal:
           path: spec.strategy.type
           value: RollingUpdate
+  - it: should override user strategy to Recreate for SQLite
+    template: deployment.yaml
+    set:
+      mode: deployment
+      config.database.type: sqlite
+      strategy:
+        type: RollingUpdate
+    asserts:
+      - equal:
+          path: spec.strategy.type
+          value: Recreate


kalbasit:
* **#1012** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/1012?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/1012?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kalbasit:
Successfully created backport PR for `release-0.9`:
- #1015

PR #1016, opened 2026-03-02
docs: expand trusted-users security documentation

### LINKED ISSUE #1011: docs: don't recommend adding `trusted-users`
In the docs under [Client Setup](https://docs.ncps.dev/user-guide/usage/client-setup#trusted-users) I found a section about extending the `trusted-users` setting.
Please add some kind of warning, like done in [nix.dev](https://nix.dev/manual/nix/2.33/command-ref/conf-file.html#conf-trusted-users), to prevent inexperienced users from undermining their systems security.
Maybe the section should also state in which cases this makes sense and in which cases there are better solutions.

kalbasit:
This change expands the documentation for the 'trusted-users' setting in
Nix to provide clearer warnings about its security implications.

Key changes:
- Replaced the warning with a more prominent caution block.
- Explicitly stated that 'trusted-users' is equivalent to root access.
- Listed specific security risks such as arbitrary binary caches and
  sandbox-path escapes.
- Added guidance on when to use this setting vs. using better
  alternatives like global configuration.
- Fixed several linting issues by adding language identifiers to code
  blocks and correcting list spacing.

fixes #1011

### DIFF
--- a/docs/docs/User Guide/Usage/Client Setup.md
+++ b/docs/docs/User Guide/Usage/Client Setup.md
@@ -173,7 +173,7 @@ trusted-users = root youruser
 ```
 
 > [!CAUTION]
-> **Adding a user to** `**trusted-users**` **is equivalent to giving that user root access to the system.**
+> **Adding a user to** `trusted-users` **is equivalent to giving that user root access to the system.**
 >
 > Trusted users can:
 >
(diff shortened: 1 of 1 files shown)


kalbasit:
* **#1016** <a href="https://app.graphite.com/github/pr/kalbasit/ncps/1016?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kalbasit/ncps/1016?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1172, opened 2026-05-14
fix: make lazy chunking opt-in (default: false)

kalbasit:
Lazy chunking was made the default in #1099, but enabling it
silently on upgrade is a behavior change: it starts background
workers, a cleanup cron job, and delays compressed NAR deletion
without user consent. Revert the default to false so users must
explicitly opt in.

Changes:
- pkg/ncps/serve.go: Value true->false, fix usage string
- charts/ncps/values.yaml: lazyChunkingEnabled true->false
- charts/ncps/tests/configmap_test.yaml: fix default assertion
- docs: update all references from default true to false


kalbasit:

- #1172 ◀

PR #1176, opened 2026-05-17
feat: add --verify-content flag to fsck

kalbasit:
Nix build errors like 'bad archive: input doesn't look like a
Nix archive' can be caused by bit-rot in CDC chunks that are
served without any content verification. The existing fsck only
stat-checks that chunk files exist — it never reads or hashes
their content.

Add an opt-in --verify-content flag that:
- Reads and decompresses every chunk via GetChunk, hashes the
  output with SHA-256, and compares it against the stored BLAKE3
  key (narFilesWithCorruptChunks).
- For NARs that pass per-chunk checks, assembles the full stream
  and compares the SHA-256 against the narinfo NarHash field
  (narFilesWithHashMismatch).
- Repairs both categories with the same cascade as broken CDC
  NARs (nar_file + orphaned narinfo + orphaned chunks).
- Shows the two new rows in the fsck summary only when
  --verify-content is active.

The flag is opt-in because reading all chunk bytes is O(total
NAR bytes) of I/O and zstd decompression. Combine with
--verified-since to limit scope on large caches.

Helm chart: adds fsck.verifyContent value (default false) and
the corresponding --verify-content arg in the CronJob template.
Docs: new usage section, flags reference row, and repair table
entries.


kalbasit:

- #1176 ◀

PR #1188, opened 2026-05-20
feat(migrate): ncps migrate up + adoption + Schema.Create path

kalbasit:
Implements §9 of the migrate-to-ent-and-atlas change — the runtime
migration system that ties together the Ent schemas (§4), translated
dbmate-era migrations (§7), generate-migrations tool (§6), and
ent_baseline bridge (§8) into an operator-facing CLI.

New package pkg/database/migrate/ exposes:

- Up(ctx, Options) error: probes adoption state, dispatches to the
  appropriate path (fresh install / dbmate adoption / goose hand-off),
  then runs incremental migrations via goose.
- DryRun(ctx, Options) (Plan, error): same probe but no DDL; returns
  a Plan describing what Up would do.
- Down(ctx, Options) error: returns ErrDownNotSupported with a
  pointer to the expand-contract recipe (forward-only policy per
  design D10).

State detection (state.go) distinguishes six cases per design D6:
empty, dbmate-shape, adopted (goose-shape), MySQL S4 (mid-adoption,
backup table only), MySQL S5 (mid-adoption, both tables, new table
has goose shape), and impossible-S6 (manual intervention required).

Fresh-install (fresh.go) takes the Schema.Create fast path:
- Ent's Schema.Create produces the entire end-state schema in one
  shot.
- Goose's database.NewStore creates the schema_migrations table in
  goose's canonical shape (avoiding shape conflicts with goose's
  own CREATE).
- Insert version 0 (the goose sentinel) + every embedded version
  stamp marked is_applied=true.
- Goose's later Up sees nothing pending and is a no-op.

A process-wide mutex serialises Schema.Create across goroutines
because Ent mutates the package-level migrate.Tables slice during
Create. Production runs a single ncps process so this is invisible;
the parallel test suite tripped it under -race.

Adoption (adopt.go) is dialect-asymmetric:

- SQLite + Postgres: single transaction. CREATE TEMP, DROP old,
  CREATE new in goose shape, INSERT sentinel + preserved versions,
  verify row-count parity (preCount+1 incl. sentinel), COMMIT or
  ROLLBACK.

- MySQL: non-transactional state machine. S3 (full adoption):
  RENAME → CREATE → sentinel + copy → verify → DROP backup. S4
  (crashed after RENAME): resume from CREATE. S5 (crashed
  mid-copy): row-count verify; on mismatch TRUNCATE and re-copy,
  then DROP backup. S6 (impossible): ErrImpossibleState.

[...]

PR #1187, opened 2026-05-20
feat(migrations): ent_baseline bridge + schema-equivalence test

kalbasit:
Implements §8 of the migrate-to-ent-and-atlas change. Closes the gap
between the §7 translated dbmate-era migrations and Ent's expected
schema (§4) with a per-dialect bridge migration, and gates the
equivalence permanently via a CI-runnable test.

Bridge migrations (one per dialect, all sorted after the last
dbmate-era file):

- migrations/sqlite/20260520212039_ent_baseline.sql (Atlas-auto):
  the SQLite-canonical "recreate via temp table" pattern. Adds
  surrogate id PKs on weak entities (D10b), renames indexes to Ents
  naming convention (e.g. idx_narinfos_hash -> narinfo_hash), and
  adds Ent's named CHECK constraints.

- migrations/mysql/20260520212039_ent_baseline.sql (Atlas-auto):
  ALTER TABLE ADD COLUMN id BIGINT AUTO_INCREMENT, DROP PRIMARY KEY,
  ADD PRIMARY KEY (id), ADD UNIQUE INDEX on weak entities. Plus
  MODIFY COLUMN to varchar(255) for hashes, bigint for chunk sizes
  (signed; previously bigint unsigned), and FK constraint renames.

- migrations/postgres/ — split into two files because Atlas's diff
  engine refuses to compute the BIGSERIAL/IDENTITY reverse direction:
  - 20260520212038_postgres_serial_to_identity.sql (hand-written):
    DROP DEFAULT + ADD GENERATED BY DEFAULT AS IDENTITY for each of
    the five existing id-bearing tables, setval the new IDENTITY
    sequence past max(id), drop the orphaned BIGSERIAL sequences.
    The BIGSERIAL -> IDENTITY modernization is bundled here per
    design D6 so v1 lands with modern Postgres conventions and
    v1.x doesnt need a separate disruptive migration.
  - 20260520213017_ent_baseline.sql (Atlas-auto, after the IDENTITY
    conversion file is in place so Atlas can compute the rest):
    surrogate ids on weak entities, text -> varchar, integer ->
    bigint, timestamp -> timestamptz, FK and CHECK constraint
    renames.

Tooling:

- cmd/generate-migrations extends §6 with two improvements:
  1. ensureAtlasSum: computes and writes atlas.sum after creating
     a GooseDir, so the first run on a hand-translated directory
     (per §7) bootstraps the integrity file. atlas.sum is committed
     for all three dialects.
[...]

PR #1186, opened 2026-05-20
docs(openspec): adopt option E for dbmate→Ent reconciliation

kalbasit:
The §8 schema-equivalence test exposed real schema drift between the
§7 translated dbmate-era migrations and Ent's expected schema:

- D10b structural: weak entities need a surrogate id PK column
- Postgres dialect: dbmate emitted BIGSERIAL; Ent emits BIGINT
  GENERATED BY DEFAULT AS IDENTITY (Atlas refuses to compute the diff
  as an "unexpected attribute change")
- Cosmetic: index and CHECK constraint naming differs between dbmate
  and Ent's defaults

Refines design D6 to capture Option E (three artifacts + decision
tree):

1. Translated dbmate-era files stay in the repo as the compatibility
   history for v0.4-era installs with partial dbmate state.

2. A single bridge migration per dialect (migrations/<dialect>/<ts>_
   ent_baseline.sql) transitions the dbmate-era schema to Ent's
   expected schema. SQLite + MySQL bridges are Atlas-generated;
   Postgres is hand-written (~30 lines of ALTER TABLE ... DROP
   DEFAULT / DROP SEQUENCE / ADD GENERATED) for the deliberate
   BIGSERIAL → IDENTITY modernization. v1 is the right cutover point
   for this modernization since the migration cost is bundled with
   the larger dbmate→Ent transition.

3. Fresh installs (empty DB) skip the 14/8/8 dbmate-era files
   entirely. The adoption code in `ncps migrate up` invokes Ent's
   runtime Schema.Create(ctx, migrate.Tables...) to produce the
   end-state schema in one operation, then seeds schema_migrations
   with every version stamp from migrations/<dialect>/ marked as
   is_applied=true. Goose subsequently sees nothing pending.

The invariant making this work: migrate.Tables and the set of files
under migrations/<dialect>/ are kept in lockstep, enforced by §13 CI
checks (ent-codegen-drift-check + schema-equivalence-check).

Future incremental migrations work without code changes: existing
installs apply each new file via goose; fresh installs land directly
on the current end state via Schema.Create + seed all known versions.
A future "squash" change can delete the dbmate-era files once no
v0.x install is left in the wild.

Updates:
- design.md D6 rewritten with the three-artifact framework, the
  adoption decision tree, long-term incremental flow, and the future
  squash plan. Rejected alternatives (A, C, D) documented with their
  reasons.
- specs/database-migrations/spec.md gains four new requirements: the
[...]

PR #1185, opened 2026-05-20
feat(cmd): add generate-migrations with atlas-as-library

kalbasit:
Adds cmd/generate-migrations, the Atlas-as-library driven migration
generator for the migrate-to-ent-and-atlas change. Atlas is consumed
through ariga.io/atlas/sql/sqltool — no atlas CLI binary is required
or packaged.

Modes:

- schema-driven (default): diffs the current Ent schemas under
  ent/schema/ against the latest applied state of migrations/<dialect>/
  by replaying the embedded history (schema.ModeReplay) against a dev
  database, then writes one .sql per dialect under a shared timestamp
  prefix. SQLite uses an in-memory db; Postgres and MySQL use dev URLs
  from flags or NCPS_GEN_POSTGRES_URL / NCPS_GEN_MYSQL_URL env vars
  (defaults match nix run .#deps).

- --sql-only: writes empty goose stubs to each dialect directory. For
  data backfills and the step-3 constraint lock-in of the four-step
  NOT NULL recipe.

Flags:
  --name=NAME       required; placeholder names rejected (auto, wip,
                    tmp, todo, temp, test, empty, whitespace)
  --sql-only        switch to stub mode
  --root=PATH       repo root (contains migrations/); defaults to .
  --postgres-url    Postgres dev DSN (or NCPS_GEN_POSTGRES_URL env)
  --mysql-url       MySQL dev DSN (or NCPS_GEN_MYSQL_URL env)

Wired into Taskfile.dist.yml as 'task migrations:gen NAME=...' and
'task migrations:sql NAME=...' (added in §2).

Smoke tests:
- TestSQLOnlyEmitsThreeDialects: --sql-only produces exactly 3 .sql
  files (one per dialect) under a single shared timestamp prefix.
- TestNameValidation: 8 reject cases and 2 accept cases.

Schema-driven roundtrip (running the generator against the current
state produces zero new files) is enforced in §8 as part of the
schema-equivalence golden test — that's the natural home because
it's the same proof + the same fixtures.

Completes tasks 6.1-6.8 of the migrate-to-ent-and-atlas change.
6.6 (atlas.sum) is delegated to Atlas's sqltool.NewGooseDir which
maintains atlas.sum integrity automatically.

## Summary by CodeRabbit

* **New Features**
  * Added `generate-migrations` CLI tool for generating SQL migrations across PostgreSQL, MySQL, and SQLite.
  * Introduced `--sql-only` mode for generating migration stubs without database interaction.

* **Tests**
[...]

PR #1190, opened 2026-05-21
docs(openspec): address PR #1178 review comments

kalbasit:
docs(openspec): address PR #1178 review comments

Three review comments on design.md, all valid:

- Tracking-table name inconsistency (PRRT_kwDONV6tZs6DokBc): D5
  said `ncps_schema_versions` while D6 and the database-migrations
  spec said `schema_migrations`. The latter is the intended choice
  (preserves dbmate-era name for operator continuity). Updated D5.

- tstamp NULL → NOT NULL on the SQLite/Postgres adoption block
  (PRRT_kwDONV6tZs6DokBp): goose's actual Postgres dialect emits
  NOT NULL. Updated the design doc + adopt.go's adoptPostgresSteps
  to match.

- tstamp NULL → NOT NULL on the MySQL S3 adoption documentation
  (PRRT_kwDONV6tZs6DokBt): in modern MySQL (with
  explicit_defaults_for_timestamp=ON) the column is effectively
  NOT NULL anyway, so make the spec stricter and consistent.
  No impl change needed — createMySQLSchemaMigrations already uses
  NULL DEFAULT which functions identically.

All migrate-package tests pass under -race against all three
engines after the Postgres NOT NULL change.

fix(Taskfile): quote NAME variable to prevent shell injection

Two review comments on PR #1179 (PRRT_kwDONV6tZs6DolmP and
PRRT_kwDONV6tZs6DolmV): the `{{.NAME}}` template variable was
expanded unquoted into the shell command. Whitespace or shell
metacharacters in NAME would either misparse or open a small shell
injection surface. Wrap in double quotes.

The third comment on this PR (PRRT_kwDONV6tZs6DolmN: goose missing
from go.mod) was correct at the time of the commit but has since
been organically fixed — once §7's apply_test.go imported goose,
`go mod tidy` pulled it in as a direct dep. Resolved without code
change.

test(schema-parity): lowercase introspected names consistently

Two review comments on PR #1181:

- PRRT_kwDONV6tZs6DokIU (mustListTables): the MySQL/Postgres queries
  return table names in their stored case, which can drift from the
  lowercase expectedTables list under some MySQL configs. Lowercase
  on read, matching the column-introspection pattern.

- PRRT_kwDONV6tZs6DokIX (mustListColumns / SQLite branch): SQLite's
  PRAGMA table_info returns column names as-declared; the
  MySQL/Postgres branches already lowercase. Lowercase the SQLite
  branch too for consistency.

The third comment (PRRT_kwDONV6tZs6DokIM: use assert.ElementsMatch
for strict column-set equality instead of Contains) is rejected
with rationale: the same expectedColumns fixture is consumed by
TestSchemaParity (dbmate schema, exact match would work) AND
TestEntSchemaParity (Ent schema, which legitimately adds the
surrogate `id` columns on weak entities per design D10b). Strict
ElementsMatch would fail on TestEntSchemaParity for legitimate
reasons. Contains correctly models "the documented columns must
exist; extras are allowed".

fix(ent-schema): use Nillable for all optional time fields

Two review comments on PR #1182 (PRRT_kwDONV6tZs6DomHI,
PRRT_kwDONV6tZs6DomHL): nullable time fields in the generated Ent
structs were inconsistent — NarFile.ChunkingStartedAt used
sql.NullTime (which marshals as a JSON object), NarFile.LastAccessedAt
and NarInfo.LastAccessedAt used time.Time (which can't distinguish
NULL from the zero value).

Updated the schema source (ent/schema/nar_file.go,
ent/schema/narinfo.go) to chain .Nillable() on all three fields and
removed the now-unused GoType(sql.NullTime{}) override. Regenerated
the ent client.

All three nullable time fields now use *time.Time consistently, both
for JSON-correctness and for distinguishing NULL from zero. The
sqlc.yaml override that originally forced sql.NullTime is no longer
needed (sqlc tree is removed in §12 cleanup).

Tests pass across all three engines under -race.

fix(ent-lint): skip top-level helpers and unparseable edge targets

Four review comments on PR #1183 (PRRT_kwDONV6tZs6Dont7,
PRRT_kwDONV6tZs6DonuJ, PRRT_kwDONV6tZs6DonuN, PRRT_kwDONV6tZs6DonuR):

Three of them point out that checkA1, checkA2, and checkA4 iterate
methods named Fields/Edges but don't verify the function has a
receiver. If a future ent/schema/ file ever contains a top-level
helper function with one of those names, the linter would
incorrectly attribute its body to a non-existent schema. Skip
methods whose receiverTypeName is empty.

The fourth points out that checkA4 was constructing edgeTo entries
even when typeRefName returned "" (the edge target isn't a simple
T.Type expression — e.g. a more complex Ent invocation we don't
model). The resulting empty-target entry would produce confusing
failure messages downstream. Skip the edge in that case.

ent-lint passes against the live ent/schema/ tree; the fixture-based
unit tests pass under -race.

fix(migrations): goose-correct SQLite PRAGMA + MySQL InnoDB defaults

Three review comments on PR #1184:

- PRRT_kwDONV6tZs6DojlB (HIGH): the translated SQLite migration
  20260127223000_allow_multiple_nar_representations.sql uses the
  PRAGMA foreign_keys = OFF / DROP TABLE / RENAME pattern to swap
  the nar_files table. SQLite ignores PRAGMA foreign_keys inside a
  transaction, and goose wraps migrations in a transaction by
  default — so against any data already in narinfo_nar_files, the
  DROP TABLE would fail. Added `-- +goose NO TRANSACTION` plus a
  comment explaining why. Same fix applied to the ent_baseline
  bridge migration (20260520212039_ent_baseline.sql) which also
  uses PRAGMA + table swaps.

- PRRT_kwDONV6tZs6DojlH: 20260131021850_add_chunks.sql for MySQL
  was missing the ENGINE/CHARSET/COLLATE clauses that
  20260101000000_init_schema.sql sets. Without them the tables
  would inherit server defaults that may not be InnoDB +
  utf8mb4_unicode_ci. Added the clauses to both CREATE TABLE
  statements.

- PRRT_kwDONV6tZs6DojlK: the SQLite Down migration in
  20260301000000_add_verified_at_to_nar_files.sql used a fragile
  full-table-swap pattern that required manually maintaining the
  column list. Replaced with the much simpler
  `ALTER TABLE nar_files DROP COLUMN verified_at` (supported by the
  project's go-sqlite3 v1.14.x). This also removes another
  PRAGMA-inside-transaction usage. Down migrations aren't supported
  per design D10, but the cleanup keeps the file honest.

atlas.sum regenerated for all three dialects. All tests pass under
-race across SQLite, PostgreSQL, and MySQL.

Note: the Postgres create_test_db helper function disappeared again
during the test run (this is the third time this session); reinstalled
manually before re-running. May want to investigate why the deps
init isn't sticky between runs.

fix(generate-migrations): use migrate.Tables as source of truth

Review comment on PR #1185 (PRRT_kwDONV6tZs6DonIx): the resetDevDB
helper hardcoded the list of ncps tables to drop before each
migration-generation run. If a future entity is added to ent/schema/
the hardcoded list would silently drift, leaving stale tables in
the dev DB and corrupting the diff.

Switched to enumerating migrate.Tables (already imported in this
file) and appending schema_migrations explicitly. Also restructured
the dialect switch to handle "postgres" explicitly and error on
unknown dialects, so adding a new engine to the loop won't silently
fall through to no-op behavior.

cmd/generate-migrations tests pass; the smoke test verifies the
generator still produces zero diff against the current
schemas/migrations state.

docs(openspec): address PR #1186 review comments

Five comments on the Option E docs PR:

- PRRT_kwDONV6tZs6DolgE (HIGH): the renamed D6 section introduced a
  contradiction — the legacy adoption sub-sections still said
  "Table doesn't exist → fresh install. Goose creates
  schema_migrations and runs all translated migrations from scratch"
  but Option E handles fresh installs via Schema.Create. Removed
  the contradictory "Table doesn't exist" branches from the
  SQLite/Postgres and MySQL adoption sub-sections and clarified
  that those sub-sections describe the *table-conversion mechanic*
  for the dbmate-shape branch only; the empty-DB case is handled
  by the higher-level Schema.Create path.

- PRRT_kwDONV6tZs6DolgG (HIGH): the BIGSERIAL → IDENTITY conversion
  on Postgres needs sequence sync (setval past max(id)) or the new
  IDENTITY starts at 1 and the first post-bridge INSERT collides.
  The actual hand-written
  20260520212038_postgres_serial_to_identity.sql already does this
  — the doc was just incomplete. Added the explanation.

- PRRT_kwDONV6tZs6DolgM / PRRT_kwDONV6tZs6DolgS: replaced bare
  references to `NOW()` with "the dialect-appropriate function or a
  Go-side time.Now() value" since SQLite uses CURRENT_TIMESTAMP not
  NOW().

- PRRT_kwDONV6tZs6DolgZ: noted the partial-fresh crash recovery
  scenario in tasks.md as a known edge case deferred to follow-up.
  Currently if Schema.Create succeeds but version seeding crashes,
  Detect returns ErrCorruptState. Adding a StatePartialFresh would
  improve robustness but the same DB shape can arise from genuine
  corruption / manual operator actions; auto-recovery would mask
  those. For v1, refusing to adopt is the safer default.

test(migrate): robust mysql DSN + MkdirAll in copyMigrations

Two review comments on PR #1187:

- PRRT_kwDONV6tZs6DooEC: the hand-rolled mysqlURLToDSN concatenated
  user/password/host/db into a DSN string, which breaks if the
  password contains `:` or `@`. Switched both copies of the helper
  (migrations/equivalence_test.go and
  pkg/database/migrate/migrate_test.go) to use
  mysqldriver.NewConfig, which handles escaping correctly. Also
  promotes parseTime / loc / multiStatements to proper Config
  fields when they're not already set in the URL params.

- PRRT_kwDONV6tZs6DooEF: copyMigrations in equivalence_test.go used
  os.Create directly, which would fail if a future migrations
  layout introduced subdirectories. Added MkdirAll on the target
  file's parent directory. Today the layout is flat so this is a
  no-op, but the helper is now robust to a future refactor.

Migrations + migrate-package tests pass under -race.

fix(migrate): idiomatic rollback + dynamic appTables + io.Writer plan

Six review comments on PR #1188, all valid:

- PRRT_kwDONV6tZs6DonwL (adopt.go): replace manual `rollback()` calls
  on every error path with a single `defer tx.Rollback()`. The
  deferred call is a no-op after a successful Commit, and the
  pattern is robust to future early-return paths.

- PRRT_kwDONV6tZs6DonwR + PRRT_kwDONV6tZs6DonwW (state.go): the
  hardcoded `appTables` list would drift from the Ent schema as
  entities are added. Switched to populating it at init() from
  `entmigrate.Tables`, matching the same approach we applied to
  cmd/generate-migrations in PR #1185's fix.

- PRRT_kwDONV6tZs6Donwa + PRRT_kwDONV6tZs6Donwc +
  PRRT_kwDONV6tZs6Donwd (ncps/migrate.go): `printPlan` hardcoded
  `os.Stdout`. Updated to take an `io.Writer` (testable + respects
  the cli.Command's configured Writer) and the caller passes
  `cmd.Writer` (falling back to `os.Stdout` if unset).

All migrate-package + ncps tests pass under -race across all three
engines.


kalbasit:

- #1178
    - #1179
        - #1180
            - #1181
                - #1182
                    - #1183
                        - #1184
                            - #1185
                                - #1186
                                    - #1187
                                        - #1188
                                            - #1190 ◀
                                                - #1191
                                                    - #1192
                                                        - #1193
                                                            - #1194
                                                                - #1195
                                                                    - #1196
                                                                        - #1197
                                                                            - #1198
                                                                                - #1199
                                                                                    - #1200
                                                                                        - #1201
                                                                                            - #1202
                                                                                                - #1203
                                                                                                    - #1204
                                                                                                        - #1205
                                                                                                            - #1206
                                                                                                                - #1207
                                                                                                                    - #1208
                                                                                                                        - #1209
                                                                                                                            - #1210
                                                                                                                                - #1211
                                                                                                                                    - #1212
                                                                                                                                        - #1213
                                                                                                                                            - #1214
                                                                                                                                                - #1215
                                                                                                                                                    - #1216
                                                                                                                                                        - #1217
                                                                                                                                                            - #1218
                                                                                                                                                                - #1219
                                                                                                                                                                    - #1220
                                                                                                                                                                        - #1221
                                                                                                                                                                            - #1222
                                                                                                                                                                                - #1223
                                                                                                                                                                                    - #1224
                                                                                                                                                                                        - #1225
                                                                                                                                                                                            - #1226
                                                                                                                                                                                                - #1227
                                                                                                                                                                                                    - #1228
                                                                                                                                                                                                        - #1229
                                                                                                                                                                                                            - #1231
                                                                                                                                                                                                                - #1232
                                                                                                                                                                                                                    - #1233
                                                                                                                                                                                                                        - #1234

PR #1194, opened 2026-05-21
feat(cache): add withEntTransaction infrastructure helper

kalbasit:
Adds the Ent-backed counterpart to *Cache.withTransaction so
upcoming §11.2-§11.5 caller-migration commits can convert
qtx-style closures one at a time without rewriting the
deadlock-retry policy each time.

The helper:

- delegates the lifecycle (begin → fn(tx) → commit / rollback
  / panic-rollback) to *database.Client.WithTransaction added
  in §10
- preserves the existing retry-on-deadlock loop verbatim (5
  attempts, 50ms initial back-off doubling per try, ctx-aware
  pacing, deadlock detection via database.IsDeadlockError) so
  conversions are behavior-preserving
- is annotated //nolint:unused because no call site is
  converted yet; the annotation will come off as soon as the
  first qtx-style block migrates over

Once every withTransaction call site moves off the Querier
closure shape, the legacy withTransaction (and the underlying
executeTransaction) become dead code and are removed in §12.

No behavior change. Lint clean, full pkg/cache -race test
suite passes against SQLite + Postgres + MySQL + MinIO + Redis.

PR #1193, opened 2026-05-21
feat(cache): convert non-tx GetNarInfo* lookups to Ent

kalbasit:
§11.2 partial: convert the GetNarInfo* call sites in
pkg/cache/cache.go that don't sit inside the legacy
*sql.Tx-based transaction wrapper and don't cascade into
sqlc-row-shape consumers.

Converted (5 sites):

- `GetNarInfoCount` → `dbClient.Ent().NarInfo.Query().Count(ctx)`
  (and the metrics observer now casts the int result to int64).
- `GetNarInfoURLByNarFileHash` (a 3-table join through
  narinfo_nar_files → nar_files) → an `entnarinfo.HasNarInfoNarFilesWith
  (entnarinfonarfile.HasNarFileWith(...))` predicate. The result
  shape changes from `sql.NullString` to a `*ent.NarInfo` whose
  `URL *string` field is read directly.
- `GetNarInfoHashByNarURL` → `Query().Where(URL(...)).First(ctx)`,
  with `ni.Hash` taking the place of the prior string return.
- `GetNarInfoHashesByURL` → `Query().Where(URL(...)).All(ctx)`,
  iterating `ni.Hash` instead of the prior `[]string` slice.
- `deleteNarInfoFromStore`'s existence-check + `PinClosure`'s
  pre-pin existence check both use
  `Query().Where(HashEQ(hash)).First(ctx)` with
  `ent.IsNotFound(err)` replacing `database.IsNotFoundError`.

Deferred (still on the original `c.db`/`qtx` API):

- `CheckAndFixNarInfo` and `lookupNarInfoForFix` (line 4282
  area) cascade into `niRow.URL.Valid`/`niRow.FileSize.Int64`
  consumers and into `checkAndFixNarInfoNoCompression`. That
  conversion needs to flow row shape changes through several
  helpers; left for a follow-up commit.
- The closure-pinning narinfo + references lookup (lines 6897,
  6907) reads `ni.ID` and feeds it back into
  `GetNarInfoReferences`. Same row-shape cascade — follow-up.
- Every `qtx.GetNarInfo*` site inside `c.executeTransaction`
  blocks (~6 sites) needs the transaction wrapper to first
  switch from `func(qtx database.Querier) error` to
  `func(tx *ent.Tx) error`. That's an infrastructure change
  pre-requisite, also follow-up.

Imports: adds `ent`, `entnarfile`, `entnarinfo`,
`entnarinfonarfile`. The legacy `database` import stays — most
[...]

PR #1192, opened 2026-05-21
feat(cache): thread *database.Client into Cache

kalbasit:
§11.1 of migrate-to-ent-and-atlas. Lays the threading plumbing
for the Querier → Ent.Client migration without converting any
call sites yet — that work is split across §11.2-§11.7.

Changes:

- `pkg/cache/cache.Cache` gains a `dbClient *database.Client`
  field alongside the existing `db database.Querier`. Both share
  a single `*sql.DB` (one connection pool). The dbClient field
  is unused for now; subsequent §11 commits will progressively
  replace `c.db.<Method>(…)` call sites with `c.dbClient.Ent().…`.
- `cache.New` and `pkg/ncps.createCache` accept the dbClient as
  a new parameter; `pkg/ncps/serve.createDatabaseQuerier` now
  also constructs the Client (via `database.NewClient(db.DB(),
  detected_type)`) and returns it alongside the Querier.
- `testhelper.Setup{SQLite,Postgres,MySQL}` likewise yield the
  dbClient; every Setup caller (cache, cache/healthcheck, cache
  distributed, server, ncps test files) destructures it
  accordingly. Wrapper helpers — `newTestCache`,
  `setupTestComponents`, the `cacheFactory` type, the
  `distributedDBFactory` type, `setup{SQLite,Postgres,MySQL}Factory`,
  and `runTestsWithDB` in `cache_internal_test.go` — gain the
  Client in their return tuples / parameter lists. Tests that
  don't need the Client discard it with `_`.
- Sites that open SQLite directly (without testhelper) construct
  the Client inline next to `database.Open`.

Why now and not later: every test file calling `cache.New`
already breaks the moment the constructor changes shape, so the
threading has to land atomically. The conversion of actual
Querier method calls is deferred to §11.2-§11.7 to keep each
commit reviewable.

No behaviour change. Full `go test -race ./...` passes against
SQLite + Postgres + MySQL + MinIO + Redis.

PR #1191, opened 2026-05-21
feat(database): introduce *Client surface

kalbasit:
Lays the foundation for §10 of migrate-to-ent-and-atlas — replacing
the sqlc-generated `Querier` interface with an Ent-backed
`*database.Client`.

What lands:

- `pkg/database/client.go` defines `*Client` (wraps `*ent.Client`,
  the underlying `*sql.DB`, and the dialect Type) plus a small set
  of accessors (`Ent`, `DB`, `Type`, `Close`).
- `WithTransaction(ctx, name, fn func(*ent.Tx) error) error` is the
  Ent-native replacement for the legacy
  `*Cache.executeTransaction` helper: begin → fn → commit on
  success, rollback on error, defer-rollback on panic (with a new
  `ErrTransactionPanic` sentinel callers can `errors.Is` against).
  Operation name flows into every wrapped lifecycle error so logs
  and telemetry can attribute failures back to the originating
  call.
- `NewClient(*sql.DB, Type)` constructs the wrapper from a
  caller-owned `*sql.DB`. The migrate package already opens its
  own `*sql.DB` for goose, so this keeps that lifecycle untouched.
- `EntDialectFor(Type)` exports the existing local mapping so the
  migrate package and future callers share one source of truth.
- `pkg/database/client_test.go` exercises commit, rollback,
  panic-rollback, begin-error wrap, accessors, Close, and the
  dialect mapping.

Why split §10.1 from the `Open` signature swap:

The task list literally says "Replace `database.Querier` return type
with `*database.Client` ... in `database.Open`", but 22 caller files
hold `database.Querier`. Swapping `Open` here would leave the build
red until §11 (caller migration) lands. The autonomous-drive rule
this stack follows ("make sure tests are passing" after every group)
forbids that. §10 therefore introduces the new surface in parallel;
§11 will rewire `Open` and migrate all callers in one coherent
commit.

§10.3 — `DetectFromDatabaseURL` and `PoolConfig` — is satisfied by
inspection: nothing in this commit touches them, and their existing
tests (`url_test.go`, `url_internal_test.go`, `backends_test.go`)
still pass under -race against all three engines.

PR #1195, opened 2026-05-21
refactor(database): share Schema.Create mutex across callers

kalbasit:
Ent's migrate.Tables is a package-level slice that Atlas's
setupTables mutates during Schema.Create. Anyone in the process
calling Schema.Create concurrently with a different mutex would
race on that slice, and the race detector would catch it the
moment a second caller is added.

Until now there was a single caller (pkg/database/migrate's
fresh-install path) plus an unrelated test helper in
pkg/database that used its own private sync.Mutex. They didn't
contend because the second never ran in the same test binary
as the first. The upcoming testhelper.CreateMigrateDatabase
rewrite (next commit) will add a third caller, at which point
the two mutexes would no longer be enough.

This commit consolidates them by lifting the mutex into
pkg/database as the exported `SchemaCreateMu`. Every code path
that calls Schema.Create acquires it. The lifted mutex lives
next to `Client` because both relate to Ent client lifecycle
and pkg/database is the natural home; pkg/database/migrate
already imports pkg/database so the dependency direction stays
clean.

No behaviour change beyond the test-time race elimination.

### DIFF
--- a/pkg/database/client.go
+++ b/pkg/database/client.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"sync"
 
 	"entgo.io/ent/dialect"
 
@@ -12,6 +13,16 @@ import (
 
 	"github.com/kalbasit/ncps/ent"
 )
+
+// SchemaCreateMu serialises Ent's Schema.Create across the process.
+// Ent mutates the package-level `migrate.Tables` slice during
+// Schema.Create; without a global lock, concurrent goroutines race on
+// that slice. Every code path that calls Schema.Create (fresh-install
+// in pkg/database/migrate, schema-bootstrap helpers in tests) must
+// acquire this mutex.
+//
+//nolint:gochecknoglobals // process-wide concurrency guard.
+var SchemaCreateMu sync.Mutex
 
 // ErrUnknownDialect is returned when a database.Type cannot be mapped to
 // an ent dialect string. Mirrors the sentinel in pkg/database/migrate.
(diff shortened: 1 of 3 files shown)


kalbasit:

- #1178
    - #1179
        - #1180
            - #1181
                - #1182
                    - #1183
                        - #1184
                            - #1185
                                - #1186
                                    - #1187
                                        - #1188
                                            - #1190
                                                - #1191
                                                    - #1192
                                                        - #1193
                                                            - #1194
                                                                - #1195 ◀
                                                                    - #1196
                                                                        - #1197
                                                                            - #1198
                                                                                - #1199
                                                                                    - #1200
                                                                                        - #1201
                                                                                            - #1202
                                                                                                - #1203
                                                                                                    - #1204
                                                                                                        - #1205
                                                                                                            - #1206
                                                                                                                - #1207
                                                                                                                    - #1208
                                                                                                                        - #1209
                                                                                                                            - #1210
                                                                                                                                - #1211
                                                                                                                                    - #1212
                                                                                                                                        - #1213
                                                                                                                                            - #1214
                                                                                                                                                - #1215
                                                                                                                                                    - #1216
                                                                                                                                                        - #1217
                                                                                                                                                            - #1218
                                                                                                                                                                - #1219
                                                                                                                                                                    - #1220
                                                                                                                                                                        - #1221
                                                                                                                                                                            - #1222
                                                                                                                                                                                - #1223
                                                                                                                                                                                    - #1224
                                                                                                                                                                                        - #1225
                                                                                                                                                                                            - #1226
                                                                                                                                                                                                - #1227
                                                                                                                                                                                                    - #1228
                                                                                                                                                                                                        - #1229
                                                                                                                                                                                                            - #1231
                                                                                                                                                                                                                - #1232
                                                                                                                                                                                                                    - #1233
                                                                                                                                                                                                                        - #1234

PR #1184, opened 2026-05-20
feat(migrations): translate 30 dbmate files to goose format

kalbasit:
Translates the existing dbmate-era SQL migration files to goose format,
preserving the timestamp prefix and DDL body verbatim. Each file gets
the header rewritten:

  -- migrate:up   →  -- +goose Up
  -- migrate:down →  -- +goose Down

Layout:

  migrations/
  ├── sqlite/      (14 files, 20241210054814 → 20260318053903)
  ├── postgres/    (8 files,  20260101000000 → 20260318053904)
  ├── mysql/       (8 files,  20260101000000 → 20260318053904)
  └── migrations.go (//go:embed FS embed.FS)

Per design D6, dialect histories are preserved as-is: SQLite has
14 entries because of four early SQLite-only migrations from 2024-2025;
Postgres and MySQL converge at the unified 20260101000000_init_schema.

The single embed.FS variable (migrations.FS) is consumed by goose via
fs.Sub(migrations.FS, dialect) at runtime — §9 will wire this into
ncps migrate up. atlas.sum files are not produced here; they are
written by cmd/generate-migrations on its first invocation (§6).

Smoke tests:
- migrations.TestEmbedFS asserts per-dialect file counts and the
  presence of -- +goose Up markers in every file, plus the absence of
  any dbmate-era -- migrate:up/down markers.
- migrations.TestGooseApplySQLite proves all 14 SQLite migrations
  apply cleanly via goose.NewProvider against a fresh database, that
  schema_migrations is populated with is_applied=1, and that the
  load-bearing tables exist after apply.

Full multi-engine schema-equivalence (post-apply schema matches Ent
expectation under atlas migrate diff) is gated separately in §8 once
cmd/generate-migrations is in place.

Completes tasks 7.1-7.3, 7.5, 7.6 of the migrate-to-ent-and-atlas
change. 7.4 (atlas.sum generation) is deferred to §6.

PR #1183, opened 2026-05-20
feat(ent-lint): enforce A1/A2/A4 codegen invariants

kalbasit:
Adds cmd/ent-lint, a Go AST linter that catches three of Ent codegens
silently-dropped patterns:

- A1: field-level entsql.Check(...) annotations are dropped — CHECKs
  MUST live on the schemas table-level Annotations() method.
- A2: entsql.OnDelete(...) on edge.From(...) is dropped — OnDelete
  MUST live on edge.To(...).
- A4: every edge.To(name, T.Type) needs a reciprocal
  edge.From(...).Ref(name) on the target schema, otherwise Ent
  fabricates a phantom FK column on the target.

Implementation walks ent/schema/*.go via go/parser and go/ast,
attributing each Edges() / Fields() method to its declaring schema by
receiver type so multi-schema files are handled correctly.

Output is checklist-formatted ([PASS] / [FAIL] + invariant id +
file:line + detail); binary exits non-zero on any [FAIL].

Tests:
- cmd/ent-lint/main_test.go::TestEntLint exercises 6 fixtures
  (good/bad pair per invariant) under cmd/ent-lint/testdata/.
- TestEntLintRealSchemas pins ncpss own ent/schema/ tree as
  invariant-clean.

Tracked but not yet implemented (see tasks.md §5.5/5.7-5.10):
A3 (Unique on edge-bound FK), A5 (_ciphertext .Sensitive()),
snake_case enum-type, expand-contract migration ban, CHECK presence
cross-validation against generated SQL. These remain as follow-up
commits on the stack.

Completes tasks 5.1-5.4, 5.6, 5.11-5.13 of the migrate-to-ent-and-atlas
change.

PR #1182, opened 2026-05-20
feat(ent): author 9 schemas, mixin, and parity tests

kalbasit:
Adds the foundational Ent schemas for the migrate-to-ent-and-atlas
change:

- internal/entmixin/Timestamps mixin contributes created_at +
  updated_at with DB-level DEFAULT CURRENT_TIMESTAMP (declared via
  entsql.Default so raw-SQL inserts succeed).
- ent/schema/ contains nine schemas — ConfigEntry, NarInfo,
  NarInfoReference, NarInfoSignature, NarFile, NarInfoNarFile, Chunk,
  NarFileChunk, PinnedClosure — matching the dbmate-era tables.
- The generated ent/ tree (84 files) is committed; CI verifies it is
  up-to-date via task ent:check (wired in a later commit).

Design deviation noted in design.md D10b: the four weak/join entities
(narinfo_references, narinfo_signatures, nar_file_chunks,
narinfo_nar_files) each gain a surrogate id PRIMARY KEY because Ents
codegen does not robustly support composite-PK entities. The original
composite uniqueness is preserved via UNIQUE INDEX. Behaviour parity
(FK cascade, uniqueness, ordered chunk retrieval) is intact.

The Config Go type is renamed ConfigEntry because "Config" collides
with Ents predeclared identifier; the on-disk table is pinned to
"config" via entsql.Annotation{Table: ...}. NarInfo similarly pins
"narinfos" to defeat Ents pluralisation to "nar_infos".

New tests:
- pkg/database/ent_schema_parity_test.go: TestEntSchemaParity drives
  migrate.Tables through Ents Schema.Create against an empty sqlite
  db and runs the schema-parity assertions; passes.
- pkg/database/schema_parity_test.go: extended to cover pinned_closures
  (added expected table, expected columns, UNIQUE pinned_closures.hash
  rejection).

All schema-parity tests pass against all three engines (SQLite,
PostgreSQL, MySQL).

Completes tasks 4.1-4.11 of the migrate-to-ent-and-atlas change.

PR #1181, opened 2026-05-20
test: add schema-parity TDD baseline for all 3 engines

kalbasit:
Introduces pkg/database/schema_parity_test.go as the regression bar
for the migrate-to-ent-and-atlas change. TestSchemaParity runs 14
assertions per engine (SQLite, PostgreSQL, MySQL/MariaDB) — 42 total
— against a dbmate-migrated database:

- All 8 documented tables exist (config, narinfos, narinfo_references,
  narinfo_signatures, nar_files, narinfo_nar_files, chunks,
  nar_file_chunks).
- Every documented column on every table exists.
- All 4 declared CHECK constraints reject negative values
  (narinfos.file_size, narinfos.nar_size, chunks.size,
  chunks.compressed_size).
- All 4 declared UNIQUE constraints reject duplicates
  (narinfos.hash, chunks.hash, config.key,
  nar_files (hash, compression, query)).
- All 4 declared ON DELETE CASCADE FKs cascade correctly
  (narinfo_references, narinfo_signatures, narinfo_nar_files via
  narinfo, nar_file_chunks via nar_file).

The test layer is dialect-aware: rewriteForDialect translates the
double-quoted-identifier, ?-templated SQL to Postgres ($1,$2,...)
and MySQL (backtick identifiers) at exec time. Introspection of
table/column lists uses sqlite_master, information_schema, and
PRAGMA table_info per dialect.

This test must pass unchanged after the Ent baseline and translated
migrations land in §4/§7 — proving the new system reproduces the
current schema exactly.

Completes tasks 3.1-3.3 of the migrate-to-ent-and-atlas change.

PR #1196, opened 2026-05-21
fix(ent): preserve CURRENT_TIMESTAMP default on last_accessed_at

kalbasit:
Ent's `field.Time("last_accessed_at").Default(time.Now)` only
sets the value in Go code at create-builder time — it doesn't
emit a DB-level DEFAULT. The §8 ent_baseline bridge therefore
recreated `narinfos.last_accessed_at` and
`nar_files.last_accessed_at` as plain nullable columns, silently
dropping the `DEFAULT CURRENT_TIMESTAMP` that the dbmate-era
schema had.

Real consequence: any caller that INSERTs into narinfos /
nar_files via raw SQL (e.g. every sqlc-generated query in
flight during §11) now writes NULL into last_accessed_at,
breaking LRU eviction (which reads that column) and the
contract tests that assert
`nims[0].LastAccessedAt.Time == nims[0].CreatedAt` on insert.
The bug was masked while tests used dbmate (preserving the
default); the upcoming testhelper rewrite to migrate.Up exposed
it.

Fix:

- Add `entsql.Annotation{DefaultExpr: "CURRENT_TIMESTAMP"}` to
  the `last_accessed_at` field of both `NarInfo` and `NarFile`
  Ent schemas. Ent now emits the DB-level default; Schema.Create
  and Atlas-generated migrations both pick it up.
- Regenerate `ent/migrate/schema.go` via `go generate ./ent`.
- Regenerate the bridge migration via `go run
  ./cmd/generate-migrations -name ent_baseline`, which replaces
  the three previously-committed bridge files with refreshed
  ones whose CREATE TABLE statements include
  `last_accessed_at datetime NULL DEFAULT (CURRENT_TIMESTAMP)`
  (and the dialect equivalents). atlas.sum is regenerated by
  the same tool.

No behaviour change to anything that was already going through
Ent's create builders (those still set time.Now in Go before
INSERT). Restores parity with the dbmate-era schema for the
raw-SQL callers that still exist during §11.

Verified by `go test -race ./pkg/database/migrate/...` and
`./pkg/database/...` against SQLite + Postgres + MySQL.


kalbasit:

- #1178
    - #1179
        - #1180
            - #1181
                - #1182
                    - #1183
                        - #1184
                            - #1185
                                - #1186
                                    - #1187
                                        - #1188
                                            - #1190
                                                - #1191
                                                    - #1192
                                                        - #1193
                                                            - #1194
                                                                - #1195
                                                                    - #1196 ◀
                                                                        - #1197
                                                                            - #1198
                                                                                - #1199
                                                                                    - #1200
                                                                                        - #1201
                                                                                            - #1202
                                                                                                - #1203
                                                                                                    - #1204
                                                                                                        - #1205
                                                                                                            - #1206
                                                                                                                - #1207
                                                                                                                    - #1208
                                                                                                                        - #1209
                                                                                                                            - #1210
                                                                                                                                - #1211
                                                                                                                                    - #1212
                                                                                                                                        - #1213
                                                                                                                                            - #1214
                                                                                                                                                - #1215
                                                                                                                                                    - #1216
                                                                                                                                                        - #1217
                                                                                                                                                            - #1218
                                                                                                                                                                - #1219
                                                                                                                                                                    - #1220
                                                                                                                                                                        - #1221
                                                                                                                                                                            - #1222
                                                                                                                                                                                - #1223
                                                                                                                                                                                    - #1224
                                                                                                                                                                                        - #1225
                                                                                                                                                                                            - #1226
                                                                                                                                                                                                - #1227
                                                                                                                                                                                                    - #1228
                                                                                                                                                                                                        - #1229
                                                                                                                                                                                                            - #1231
                                                                                                                                                                                                                - #1232
                                                                                                                                                                                                                    - #1233
                                                                                                                                                                                                                        - #1234

PR #1180, opened 2026-05-20
docs: validate mysql ent+atlas+goose pipeline

kalbasit:
The §1 MySQL spike (now removed) validated that ent v0.14.6 +
ariga.io/atlas v0.36.2-pre + pressly/goose/v3 v3.27.1 work end-to-end
against MariaDB 11.4.8:

- dialect.MySQL produces correct MariaDB DDL (bigint, varchar,
  backtick-quoted identifiers, utf8mb4 charset, AUTO_INCREMENT PK).
- sqltool.GooseFormatter emits valid MySQL syntax with proper
  -- +goose Up / -- +goose Down markers and reverse DROP TABLE
  ordering.
- goose.DialectMySQL applies the baseline cleanly via
  goose.NewProvider(..., WithTableName("schema_migrations")).Up(ctx).
- All schema features survive: CHECK constraints, ON DELETE CASCADE
  foreign keys, UNIQUE indexes, named indexes — verified via
  information_schema queries.

Decision gate result: proceed with the pipeline as planned in §6 and
§9. No custom Goose formatter is required for MySQL.

Updates design.md D7 with the validation evidence. The throwaway
spike (cmd/spike-mysql/ and ent/schema/spike_*.go) is deleted; this
commit captures only the documentation update.

Completes task 1.1-1.4 of the migrate-to-ent-and-atlas change.

PR #1179, opened 2026-05-20
feat: add ent, atlas, goose deps and go-task workflow

kalbasit:
Establishes the foundation for the migrate-to-ent-and-atlas change:

- Add direct module requires for entgo.io/ent v0.14.6,
  ariga.io/atlas v0.36.2-pre, and github.com/pressly/goose/v3 v3.27.1
  (they appear as indirect in this PR because no Go file imports them
  yet; subsequent PRs in the stack will pull them to direct).
- Pin entgo.io/ent/cmd/ent in the tool() directive so go generate
  resolves the Ent codegen CLI reproducibly without a separate install.
- Add go-task to nix/dev-packages.nix (dev shell and docker-dev share
  this list; go-task is dev-only, runtime images are unaffected).
- Add Taskfile.dist.yml at the repo root defining ent:generate,
  ent:lint, ent:check, migrations:gen, and migrations:sql per the
  design's D8 workflow surface.
- Add ent/generate.go marker so go generate ./ent/... discovers the
  schema directory.

Completes tasks 2.1-2.6 of the migrate-to-ent-and-atlas change.

PR #1178, opened 2026-05-20
docs: plan ent+atlas migration via openspec

kalbasit:
Adds the migrate-to-ent-and-atlas openspec change describing a move
from sqlc + sqlc-multi-db + dbmate to entgo.io/ent for schema-as-code,
ariga.io/atlas (as a Go library) for declarative migration generation,
pressly/goose for the runtime applier, and a custom cmd/ent-lint for
five codegen invariants plus the expand-contract policy. The proposal,
design, four capability specs (database-orm, database-migrations,
ent-schema-lint, data-model delta), and a 15-group / 109-task list
live under openspec/changes/migrate-to-ent-and-atlas/.

Restructures openspec/specs/data-model/spec.md into the canonical
Requirement/Scenario format, preserving every claim from the prior
freeform document. This enables the change's data-model delta to use
proper MODIFIED operations against the now-existing Requirement
blocks rather than the ADDED + REMOVED fallback that the freeform
structure forced.

PR #1197, opened 2026-05-21
refactor(testhelper): switch to migrate.Up for schema setup

kalbasit:
Replace the dbmate exec.Command invocations in
CreateMigrateDatabase / MigratePostgresDatabase /
MigrateMySQLDatabase with the same pkg/database/migrate.Up
flow `ncps migrate up` uses in production: empty DB →
ent.Schema.Create → final schema. Three motivations:

1. The dbmate-only path produces a schema without the §10b
   surrogate `id` columns on weak entities
   (narinfo_references, narinfo_signatures,
   narinfo_nar_files, nar_file_chunks). Ent's edge traversals
   (WithReferences / WithSignatures /
   WithNarInfoNarFiles) generate SQL that selects those
   columns; tests against the dbmate schema fail with
   "no such column: narinfo_references.id". Blocked
   §11.2's cascading conversions in pkg/cache. Now unblocked.

2. Eliminates the external dbmate binary dependency from
   the test path. Tests now exercise the production migrate
   code on every run, so schema-bug regressions surface in
   unit tests instead of integration.

3. Drops the need for `runtime.Caller`-derived
   `db/migrations/<dialect>/` path lookups. The migrations
   are loaded from `migrations/` via the embedded
   `migrations.FS`, the same source of truth ncps ships in
   its binary.

Required prerequisite: the previous commit's Ent schema fix
for `last_accessed_at DEFAULT CURRENT_TIMESTAMP`. Without it,
contract tests that read `LastAccessedAt.Time` would see
NULL instead of the expected insert-time timestamp.

Full -race test suite passes against SQLite + Postgres +
MySQL with the deps stack running. Lint clean.

### DIFF
--- a/testhelper/mysql.go
+++ b/testhelper/mysql.go
@@ -3,54 +3,39 @@ package testhelper
 import (
 	"context"
 	"fmt"
+	"io/fs"
 	"net/url"
 	"os"
-	"os/exec"
-	"path/filepath"
-	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/kalbasit/ncps/migrations"
 	"github.com/kalbasit/ncps/pkg/database"
+	"github.com/kalbasit/ncps/pkg/database/migrate"
 )
 
-// MigrateMySQLDatabase will migrate the MySQL database using dbmate.
+// MigrateMySQLDatabase runs the same migrate.Up flow `ncps migrate up`
+// uses in production: empty DB → ent.Schema.Create → final schema
+// including the §10b surrogate-id columns on weak entities. Replaces
+// the legacy dbmate invocation.
 // The database URL should be in the format: mysql://user:password@host:port/database
 func MigrateMySQLDatabase(t *testing.T, dbURL string) {
 	t.Helper()
 
-	_, thisFile, _, ok := runtime.Caller(0)
-	require.True(t, ok)
+	db, err := database.Open(dbURL, nil)
+	require.NoError(t, err)
 
-	dbMigrationsDir := filepath.Join(
-		filepath.Dir(filepath.Dir(thisFile)),
-		"db",
-		"migrations",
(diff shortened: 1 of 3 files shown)


kalbasit:

- #1178
    - #1179
        - #1180
            - #1181
                - #1182
                    - #1183
                        - #1184
                            - #1185
                                - #1186
                                    - #1187
                                        - #1188
                                            - #1190
                                                - #1191
                                                    - #1192
                                                        - #1193
                                                            - #1194
                                                                - #1195
                                                                    - #1196
                                                                        - #1197 ◀
                                                                            - #1198
                                                                                - #1199
                                                                                    - #1200
                                                                                        - #1201
                                                                                            - #1202
                                                                                                - #1203
                                                                                                    - #1204
                                                                                                        - #1205
                                                                                                            - #1206
                                                                                                                - #1207
                                                                                                                    - #1208
                                                                                                                        - #1209
                                                                                                                            - #1210
                                                                                                                                - #1211
                                                                                                                                    - #1212
                                                                                                                                        - #1213
                                                                                                                                            - #1214
                                                                                                                                                - #1215
                                                                                                                                                    - #1216
                                                                                                                                                        - #1217
                                                                                                                                                            - #1218
                                                                                                                                                                - #1219
                                                                                                                                                                    - #1220
                                                                                                                                                                        - #1221
                                                                                                                                                                            - #1222
                                                                                                                                                                                - #1223
                                                                                                                                                                                    - #1224
                                                                                                                                                                                        - #1225
                                                                                                                                                                                            - #1226
                                                                                                                                                                                                - #1227
                                                                                                                                                                                                    - #1228
                                                                                                                                                                                                        - #1229
                                                                                                                                                                                                            - #1231
                                                                                                                                                                                                                - #1232
                                                                                                                                                                                                                    - #1233
                                                                                                                                                                                                                        - #1234

PR #1201, opened 2026-05-21
feat(cache): convert NarInfo/NarFile delete paths to Ent

kalbasit:
§11.3 partial. The simple delete sites in pkg/cache convert
cleanly without the UPSERT cascade that storeInDatabase
will need; landing them in a separate commit keeps each diff
reviewable.

Converted:

- `deleteNarInfoFromStore` (line 4778):
  `c.db.DeleteNarInfoByHash` →
  `c.dbClient.Ent().NarInfo.Delete().Where(HashEQ(hash)).Exec(ctx)`.
- `purgeNarInfo` (line 4076):
  switched the wrapper from `c.withTransaction(... qtx ...)`
  to `c.withEntTransaction(... tx *ent.Tx ...)`. The two
  inside-tx deletes (NarInfo + NarFile) become Ent Delete
  builders with the same predicates the sqlc query used
  (hash for NarInfo; hash + compression + query for NarFile).

Behaviour preservation:

- The cascading deletes the comment notes (references,
  signatures, narinfo_nar_files links) are still handled by
  the database-level ON DELETE CASCADE the §8 bridge
  migration set up. Ent's Delete builder doesn't need to do
  anything beyond DELETE FROM narinfos WHERE hash = ?.

Deferred to follow-up commits in §11.3:

- `storeInDatabase` and the second PutNarInfo helper (line
  4631) — both do a conditional UPSERT
  (`ON CONFLICT (hash) DO UPDATE ... WHERE url IS NULL`) plus
  bulk reference / signature inserts plus a NarFile UPSERT
  plus a NarFile<->NarInfo link. Substantial; their own
  commit.
- `DeleteNarInfoByID` inside the LRU eviction loop (line
  5535) — sits in a larger withTransaction block whose
  surrounding qtx calls belong to §11.5 (orphan cleanup).
  Leaving it qtx-style until that whole block converts as a
  unit.

`go test -race ./pkg/cache/...` passes against SQLite + Postgres
+ MySQL + Redis. Lint clean.

PR #1200, opened 2026-05-21
chore(ent): enable sql/upsert feature flag

kalbasit:
§11.3 requires expressing PutNarInfo's UPSERT semantics
(`INSERT ... ON CONFLICT (hash) DO UPDATE ... WHERE url IS NULL`)
in Ent's fluent API. The OnConflict / UpsertOne / UpsertBulk
builders are only emitted when the `sql/upsert` feature flag
is on, so this commit turns it on and regenerates.

What changed:

- `ent/generate.go`: add `--feature sql/upsert` to the codegen
  invocation. Subsequent `go generate ./ent` invocations now
  always produce OnConflict-capable builders; CI's
  `git diff --exit-code ./ent/` check stays meaningful because
  the regenerated tree is committed here.
- Every `<entity>_create.go` file gains an OnConflict method,
  a Conflict type that exposes per-field
  Update/UpdateNewValues/Ignore handlers, and the corresponding
  Upsert / UpsertOne / UpsertBulk types. Pure additions; no
  existing API changes.

No call-site code uses OnConflict yet — §11.3's caller
migrations land in the next commit. Adding the feature flag in
its own commit keeps the §11.3 conversion diff focused on the
behavioural change, not the ~4k lines of mechanical codegen.

`go build ./...` succeeds, `golangci-lint run ./...` returns
zero issues, no other commit in the stack is affected.

PR #1199, opened 2026-05-21
feat(cache): convert remaining GetNarInfo* read paths to Ent

kalbasit:
feat(cache): convert remaining GetNarInfo* read paths to Ent

§11.2 completion. Three call sites that were left unconverted
by the prior commits because they read row fields downstream
(the "row-shape cascade") and so couldn't move in isolation.
All three now run through the Ent fluent API.

- `getNarInfoFromStore` (line 3699): inside-tx GetNarInfoByHash
  + TouchNarInfo path. Switched from c.withTransaction
  (qtx-style) to c.withEntTransaction (Ent-tx style). The
  TouchNarInfo call becomes
  `tx.NarInfo.Update().Where(HashEQ).SetLastAccessedAt(time.Now())
  .Save(ctx)`, matching the same pattern populateNarInfoFromDatabase
  uses for touch.

- `CheckAndFixNarInfo` + `checkAndFixNarInfoNoCompression` (line
  4285): consumes the entire narinfo row to inspect URL,
  FileSize, FileHash. Now reads from `*ent.NarInfo` (pointer
  fields) via the new `derefInt64Ptr` helper and direct
  pointer dereferences. The two "fix to NULL" paths in
  checkAndFixNarInfoNoCompression use
  `tx.NarInfo.Update().ClearFileSize()` / `.ClearFileHash()`
  to set the columns NULL — Ent's idiomatic way to do what
  `UpdateNarInfoFileSize`/`UpdateNarInfoFileHash` did with
  sql.NullInt64/sql.NullString{Valid: false}.

- Closure-pinning BFS (line 7002): combines GetNarInfoByHash +
  GetNarInfoReferences(narinfo.ID) into one round-trip via
  `tx.NarInfo.Query().Where(HashEQ).WithReferences().Only(ctx)`,
  iterating `ni.Edges.References` directly. Avoids the surrogate-id
  lookup the legacy path needed (sqlc's GetNarInfoReferences
  takes narinfo_id, which is irrelevant under Ent's edge model).

After this commit no `c.db.GetNarInfo*` or `qtx.GetNarInfo*`
call site remains in pkg/cache — §11.2 is complete and tasks.md
gets ticked. The same downstream-row-shape pattern (deref helpers
+ Ent UpdateOne mutators) will be reused for §11.3 (PutNarInfo)
and §11.4 (GetNarFile).

`go test -race ./pkg/cache/...` passes against SQLite + Postgres +
MySQL + Redis. Lint clean.

[...]

PR #1198, opened 2026-05-21
feat(cache): convert populateNarInfoFromDatabase to Ent

kalbasit:
§11.2 continuation. Switches the cache's central narinfo-read
helper from the qtx-style Querier closures (sqlc shape: row
fields are sql.Null*, references/signatures fetched in
separate queries by ID) to the Ent fluent API (row fields are
*string/*int64 pointers, references and signatures eager-loaded
in one round-trip via `WithReferences()` / `WithSignatures()`).

Specifically:

- `populateNarInfoFromDatabase(ctx, tx *ent.Tx, hash, touch)`:
  - `tx.NarInfo.Query().Where(HashEQ(hash)).WithReferences()
    .WithSignatures().Only(ctx)` replaces three separate
    Querier calls (GetNarInfoByHash + GetNarInfoReferences +
    GetNarInfoSignatures). One query, one round-trip.
  - `ent.IsNotFound(err)` replaces `database.IsNotFoundError`.
  - sql.NullString/NullInt64 reads (`x.URL.Valid`, `x.URL.String`,
    `x.FileSize.Int64`, …) become pointer dereferences via
    `derefStringPtr`/`derefInt64Ptr` helpers introduced in this
    commit. `parseValidHashPtr` is the *string counterpart to
    `parseValidHash` and replaces it (the sql.NullString
    version was the helper's only caller, so removed).
  - Touch logic (`if time.Since(LastAccessedAt) >
    recordAgeIgnoreTouch`) now reads `*time.Time` directly and
    calls `tx.NarInfo.Update().Where(HashEQ(hash))
    .SetLastAccessedAt(time.Now()).Save(ctx)` instead of the
    sqlc-generated `TouchNarInfo`.

- The one caller `getNarInfoFromDatabase` switches from
  `c.withTransaction(ctx, op, func(qtx database.Querier))` to
  `c.withEntTransaction(ctx, op, func(tx *ent.Tx))`, the helper
  added in `feat-cache-with-ent-transaction`.

Both `parseValidHashPtr` and the deref helpers are written to
be reusable for upcoming §11 conversions (PutNarInfo,
GetNarFile, fsck) that face the same null-shape translation.

Behaviour-preserving by construction:

- The eager-load result iteration order matches sqlc's. For
  references it's narinfo_references.id ASC because the join
  uses the row order; the prior GetNarInfoReferences query
  ordered by reference's PRIMARY KEY (id) too, so the slice
  is identical.
[...]

PR #1202, opened 2026-05-21
feat(cache): convert storeInDatabase to Ent

kalbasit:
§11.3 continuation. Replaces the legacy storeInDatabase
withTransaction closure (which composed CreateNarInfo +
AddNarInfoReferences + AddNarInfoSignatures + createOrUpdateNarFile
+ LinkNarInfoToNarFile) with an Ent-fluent pipeline.

Conditional UPSERT for narinfos
-------------------------------

The legacy SQL was `INSERT … ON CONFLICT (hash) DO UPDATE …
WHERE url IS NULL` — stub records (URL NULL) get overwritten;
records with a real URL stay untouched. Ent's UpsertOne API
doesn't expose the WHERE-on-update clause directly, so the new
upsertNarInfoFromParsed helper does a read + branch:

  - row absent          → Create with the supplied fields
  - row present, URL nil → UpdateOneID overwriting every field
                           (ClearX for empty source values, to
                           match the legacy NULL semantics)
  - row present, URL set → keep existing row

This is safe because the surrounding *ent.Tx serialises the
read and write, and the cache-level write lock on
PutNarInfo prevents two same-hash callers from racing.

Two parallel helpers — applyNarInfoCreate and
applyNarInfoUpdate — keep the field-population code DRY between
the insert and stub-update branches.

Bulk reference / signature inserts
----------------------------------

addNarInfoReferences / addNarInfoSignatures use
`tx.NarInfoReference.CreateBulk(...).OnConflictColumns(...).
Ignore().Exec(ctx)`. Ignore() compiles to
`ON CONFLICT DO UPDATE SET col = col` (a no-op update) rather
than DO NOTHING; both express the legacy
`ON CONFLICT (narinfo_id, reference) DO NOTHING` behaviour but
DO NOTHING combined with RETURNING returns no rows on Postgres,
which Ent treats as `sql.ErrNoRows` — Ignore() always returns
a row, so the surrounding tx stays valid.

Resolving the conflict at the SQL layer (rather than tolerating
duplicate-key errors in Go) is required: Postgres puts the
transaction into an aborted state after any failed statement
("current transaction is aborted, commands ignored until end
of transaction block"). The previous version's
IsDuplicateKeyError tolerance broke storeInDatabase on Postgres
the moment a re-call (e.g. shared-NAR concurrent PutNarInfo)
triggered a duplicate.

NarFile upsert
--------------

createOrUpdateNarFileEnt collapses the legacy read-then-create
into a single
`tx.NarFile.Create().OnConflictColumns(hash, compression,
query).Update(SetLastAccessedAt(now)).ID(ctx)`. Atomic by
construction, no Go-level race window. Returns int (Ent's PK
type for strong entities) instead of int64; the §11.4 CDC
conversions will retire the int64-typed callers.

NarFile<->NarInfo link
----------------------

The legacy `LinkNarInfoToNarFile` was a plain INSERT … ON
CONFLICT (narinfo_id, nar_file_id) DO NOTHING. Becomes Ent's
`tx.NarInfoNarFile.Create()...OnConflictColumns(…).Ignore().
Exec(ctx)`.

Verified
--------

`go test -race ./pkg/cache/...` passes against SQLite + Postgres
+ MySQL + Redis, including the concurrency-heavy distributed
tests (PutNarInfoConcurrentSharedNar) that exercise the
conflict-resolution paths. Lint clean.

Deferred
--------

- The parallel PutNarInfo helper at line ~4631 (still uses
  the qtx pattern). It's nearly the same shape; will convert
  on top of this commit reusing the new helpers.
- `createOrUpdateNarFile` (the legacy helper) is still
  referenced from CDC code (§11.4 work). Left in place
  alongside the new createOrUpdateNarFileEnt.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -29,6 +29,8 @@ import (
 	entnarfile "github.com/kalbasit/ncps/ent/narfile"
 	entnarinfo "github.com/kalbasit/ncps/ent/narinfo"
 	entnarinfonarfile "github.com/kalbasit/ncps/ent/narinfonarfile"
+	entnarinforeference "github.com/kalbasit/ncps/ent/narinforeference"
+	entnarinfosignature "github.com/kalbasit/ncps/ent/narinfosignature"
 
 	"github.com/kalbasit/ncps/ent"
 	"github.com/kalbasit/ncps/pkg/analytics"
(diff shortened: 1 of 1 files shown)


kalbasit:

- #1178
    - #1179
        - #1180
            - #1181
                - #1182
                    - #1183
                        - #1184
                            - #1185
                                - #1186
                                    - #1187
                                        - #1188
                                            - #1190
                                                - #1191
                                                    - #1192
                                                        - #1193
                                                            - #1194
                                                                - #1195
                                                                    - #1196
                                                                        - #1197
                                                                            - #1198
                                                                                - #1199
                                                                                    - #1200
                                                                                        - #1201
                                                                                            - #1202 ◀
                                                                                                - #1203
                                                                                                    - #1204
                                                                                                        - #1205
                                                                                                            - #1206
                                                                                                                - #1207
                                                                                                                    - #1208
                                                                                                                        - #1209
                                                                                                                            - #1210
                                                                                                                                - #1211
                                                                                                                                    - #1212
                                                                                                                                        - #1213
                                                                                                                                            - #1214
                                                                                                                                                - #1215
                                                                                                                                                    - #1216
                                                                                                                                                        - #1217
                                                                                                                                                            - #1218
                                                                                                                                                                - #1219
                                                                                                                                                                    - #1220
                                                                                                                                                                        - #1221
                                                                                                                                                                            - #1222
                                                                                                                                                                                - #1223
                                                                                                                                                                                    - #1224
                                                                                                                                                                                        - #1225
                                                                                                                                                                                            - #1226
                                                                                                                                                                                                - #1227
                                                                                                                                                                                                    - #1228
                                                                                                                                                                                                        - #1229
                                                                                                                                                                                                            - #1231
                                                                                                                                                                                                                - #1232
                                                                                                                                                                                                                    - #1233
                                                                                                                                                                                                                        - #1234

PR #1205, opened 2026-05-21
feat(cache): convert CDC nar_file write paths to Ent

kalbasit:
§11.4 continuation. The in-tx NarFile write call sites that
drive CDC chunking — ensureNarFileRecord, the chunking
state machine in findOrCreateNarFileForCDC, recordChunkBatch,
storeNarWithCDC.MarkComplete, and
storeNarWithCDC.RelinkAndCleanup — all migrate to
*ent.Tx-based helpers.

Converted:

- `ensureNarFileRecord` (line ~1440): collapses the legacy
  Create-then-conditional-Update into a single
  `OnConflictColumns(hash, compression, query).Update(SetFileSize,
  SetUpdatedAt).ID(ctx)`. Idempotent and atomic.

- `findOrCreateNarFileForCDC` (line ~2149): the CDC entry
  point. Upserts with OnConflict.Update(SetUpdatedAt) for the
  touch, then `tx.NarFile.Get(ctx, id)` to inspect the
  current row state (TotalChunks, ChunkingStartedAt) — Ent's
  OnConflict only returns the ID, so a follow-up read is
  required where the legacy CreateNarFile RETURNING gave us
  the row in one round-trip. The state machine is preserved:
  total_chunks > 0 → ErrAlreadyExists; chunking_started_at
  fresh → ErrAlreadyExists; chunking_started_at stale →
  collect partial chunks via
  `tx.Chunk.Query().Where(HasNarFileLinksWith(NarFileID))`
  and delete the junction rows via
  `tx.NarFileChunk.Delete().Where(NarFileID(...))`. The lock
  bump becomes
  `tx.NarFile.UpdateOneID(...).SetChunkingStartedAt(now)`.
  Return type changes from `[]database.Chunk` to
  `[]*ent.Chunk`.

- `cleanupStaleLockChunks` (line ~2281): updated to accept
  `[]*ent.Chunk`. The staleByID lookup stays keyed on int64
  because the orphan-discovery path still uses
  `c.db.GetOrphanedChunks` (legacy Querier) which returns
  database.Chunk — that path migrates in §11.5.

- `recordChunkBatch` (line ~2347): bulk chunk persistence.
  The per-chunk Create uses OnConflictColumns(Hash).Update(
  SetUpdatedAt).ID(ctx) — matches the legacy CreateChunk's
  `ON CONFLICT(hash) DO UPDATE SET updated_at = CURRENT_TIMESTAMP`.
  The link batch uses
[...]

PR #1204, opened 2026-05-21
feat(cache): convert non-tx NarFile reads/writes to Ent

kalbasit:
§11.4 partial. The NarFile call sites in pkg/cache that sit
outside the legacy `c.withTransaction` blocks (lifecycle
queries, metric callbacks, CDC delayed-cleanup, CDC stuck-file
recovery, prefetch progress polls) all migrate cleanly to
the Ent API. Tx-bound NarFile sites in
storeNarWithCDC / findOrCreateNarFileForCDC remain on the qtx
pattern; their conversion lands in the next commit.

Converted:

- `setupMetricCallbacks` (line 706 area): three observers
  rewired to Ent.

  * `GetNarTotalSize` (SUM(file_size) with COALESCE-to-0) →
    `tx.NarFile.Query().Aggregate(ent.Sum(FieldFileSize)).
    Scan(&totals)` with a zero fallback when the table is
    empty. The same `totalSize` value is reused for the
    cache-utilization gauge, so it's lifted to an outer var.
  * `GetNarFileCount` → `Query().Count(ctx)`.
  * The narinfo count was already on Ent from §11.2; no
    change.

- `ClearNarFileChunkingStarted` (deferred chunking-failure
  cleanup, line ~1933) → `UpdateOneID(int(narFileID)).
  ClearChunkingStartedAt().SetUpdatedAt(time.Now()).Save(ctx)`.
  Matches the legacy
  `SET chunking_started_at = NULL, updated_at = CURRENT_TIMESTAMP`.

- `deleteNarFromStore`'s DB step (line ~2997) →
  `NarFile.Delete().Where(HashEQ, CompressionEQ, QueryEQ)
  .Exec(ctx)`.

- CDC delayed-cleanup (line ~5947): the legacy
  `GetOldCompressedNarFiles` did a self-join on `nar_files`
  to find compressed records whose chunked counterpart
  already exists. Ent's edge model doesn't represent the
  self-join naturally, so the conversion is a two-step query:
  1. `Select(FieldHash).Scan(&chunkedHashRows)` — gathers
     the set of hashes that already have a chunked
     (compression='none', total_chunks>0) version
  2. `Where(TotalChunksEQ(0), CompressionNEQ('none'),
     CreatedAtLT(cutoff), HashIn(chunkedHashes...))
     .All(ctx)` — picks the old compressed candidates
  Behaviour-equivalent; one extra round-trip in exchange for
  staying inside Ent's typed API. The subsequent
[...]

PR #1203, opened 2026-05-21
feat(cache): convert MigrateNarInfo + storeNarInfoInDatabase to Ent

kalbasit:
feat(cache): convert MigrateNarInfo + storeNarInfoInDatabase to Ent

§11.3 completion of the PutNarInfo write paths. The Cache-free
`storeNarInfoInDatabase` (used by `MigrateNarInfo` from the
background-migration goroutine and the CLI migrate-narinfo
command) is collapsed into the same Ent-fluent pipeline that
`Cache.storeInDatabase` adopted in the previous commit, reusing
`upsertNarInfoFromParsed`, `addNarInfoReferences`,
`addNarInfoSignatures`, and `createOrUpdateNarFileEnt`.

Key changes:

- `MigrateNarInfo` gains a `dbClient *database.Client` parameter
  alongside its existing `db database.Querier`. The Querier
  stays for the pre-lock double-check
  (`db.GetNarInfoByHash(ctx, hash)` looking for already-migrated
  rows) until §11.6 retires that surface entirely. Callers
  updated: `(*Cache).MigrateNarInfoToDatabase` already had the
  field; `pkg/ncps/migrate_narinfo.go`'s
  `migrateNarInfoCommand` extracts the Client out of
  `createDatabaseQuerier`'s second return value and passes it
  along.
- `storeNarInfoInDatabase`'s body is replaced wholesale with a
  `dbClient.WithTransaction(...)` block that runs the same Ent
  helpers as `Cache.storeInDatabase`. Same conflict-resolution
  primitives (OnConflictColumns + Ignore for references,
  signatures, and the narinfo<->nar_file link), same conditional
  read-then-update UPSERT for the narinfo row, same atomic
  Create+SetLastAccessedAt for nar_files.
- The legacy `createOrUpdateNarFile(qtx)` helper is now
  fully unused — both production call sites (the old
  storeInDatabase and storeNarInfoInDatabase) migrated to
  `createOrUpdateNarFileEnt(tx)`. Deleted.

Test infrastructure
-------------------

`testBackgroundMigrateNarInfoThunderingHerd` counts the number
of NarInfo create operations to detect "more than one
migration goroutine acquired the lock". The legacy spy
embedded `database.Querier` and counted `CreateNarInfo` calls;
since the migration path now flows through `dbClient` (Ent),
the spy never saw the conversion. The test now installs an
Ent mutation hook via
`dbClient.Ent().NarInfo.Use(func(next ent.Mutator) ent.Mutator
[...]

PR #1207, opened 2026-05-21
feat(cache): convert chunk-stream reads to Ent (junction-side)

kalbasit:
feat(cache): convert chunk-stream reads to Ent (junction-side)

§11.4 completion of the CDC streaming reads. The three
remaining Querier-based chunk-batch queries — the full-list
read in `getNarChunkedFromStore` and the two progressive
prefetch polls in `streamProgressiveChunks` — migrate to
Ent's NarFileChunk junction-side query.

Why query the junction (NarFileChunk) instead of the chunks
(Chunk):

Ent's edge-ordering generator `entchunk.ByNarFileLinks(...)`
compiles to a join + `GROUP BY chunks.id` + `ORDER BY
nar_file_chunks.chunk_index`. Postgres rejects this with
SQLSTATE 42803 ("column nar_file_chunks.chunk_index must
appear in the GROUP BY clause or be used in an aggregate
function") — a TestCDCBackends/PostgreSQL failure surfaced it
immediately. Querying the junction directly with
`WithChunk()` eager-loading avoids the implicit GROUP BY
entirely and orders by `nar_file_chunks.chunk_index` natively.
Same semantics; portable across all three engines.

Sites converted:

- `getNarChunkedFromStore` (line ~6570): replaces the legacy
  `GetChunksByNarFileID` (`SELECT … FROM chunks JOIN
  nar_file_chunks WHERE nar_file_id = ? ORDER BY chunk_index`)
  with
  `tx.NarFileChunk.Query().Where(NarFileID(...)).Order(ByChunkIndex())
  .WithChunk().All(ctx)`, then iterates `link.Edges.Chunk.Hash`
  to populate the existing `chunkHashes` slice.

- `streamProgressiveChunks` (two sites, ~6712 and ~6813):
  replaces both `GetChunksByNarFileIDFromIndex` polls with
  the same junction-side pattern plus `ChunkIndexGTE` and
  `Limit(progressivePollBatchSize)`. The loop bodies now
  iterate links and skip empty `link.Edges.Chunk` defensively
  (Ent guarantees the edge is non-nil when WithChunk is set,
  but explicit defends against future schema edits).

- `database.IsNotFoundError(err)` → `ent.IsNotFound(err)` at
  the two prefetch error sites.

- Removed the unused `entsql` import added in an earlier
  attempt at chunk-side ordering.

After this commit no `c.db.GetChunksByNarFileID*` call site
remains in pkg/cache. The §11.4 (NarFile + CDC) conversion is
[...]

PR #1208, opened 2026-05-21
fix(ncps): drop dbURL from createDatabaseQuerier error wraps

kalbasit:
fix(ncps): drop dbURL from createDatabaseQuerier error wraps

PR #1192 review: 3 sec-high comments flagged that
`createDatabaseQuerier`'s error wraps embed `dbURL` via `%q`.
Practical database URLs include credentials, leaking them into
server logs. Drop the URL from the three new wraps the §11.1
thread-through commit introduced.

Resolves:
- PR #1192 PRRT_kwDONV6tZs6DsTQ7 (line 990)
- PR #1192 PRRT_kwDONV6tZs6DsTRK (line 998)
- PR #1192 PRRT_kwDONV6tZs6DsTRQ (line 1003)

fix(cache): preserve tx debug logs + use time.Timer in retry

PR #1194 review surfaced two issues in withEntTransaction:

1. (PRRT_kwDONV6tZs6DsQzn) Restore "starting transaction" /
   "committed successfully" debug logs that the legacy
   executeTransaction emitted. Behind Debug level so noise
   stays opt-in; restores observability parity.

2. (PRRT_kwDONV6tZs6DsQzo) Replace `case <-time.After(delay):`
   inside the retry select with `time.NewTimer(delay)` +
   `t.Stop()` in the ctx-cancellation branch to avoid leaking
   the underlying timer when ctx wins the race.

Behaviour-preserving on the happy path.

fix(migrations): restore +goose NO TRANSACTION on sqlite bridge

PR #1196 (PRRT_kwDONV6tZs6DsQXZ, high): the
`-- +goose NO TRANSACTION` directive was dropped from the
SQLite ent_baseline bridge when `fix-ent-preserve-current-
timestamp-default` regenerated the migration via
`cmd/generate-migrations`.

In SQLite, `PRAGMA foreign_keys = off` is a no-op inside a
transaction. Goose wraps migrations in a transaction by
default, so without the NO TRANSACTION directive the PRAGMA
silently has no effect — and the subsequent
`DROP TABLE <parent>` / `ALTER TABLE new_X RENAME TO X` swaps
would fail in production against any database with existing
FK-referenced rows. The original §8 bridge had this directive;
the regeneration lost it because Atlas's GooseFormatter
doesn't emit it automatically.

Restored the directive plus its explanatory comment, and
regenerated `migrations/sqlite/atlas.sum` so Atlas's
checksum validator accepts the file.

Postgres and MySQL bridges don't need an equivalent — they
use real `ALTER TABLE` for column changes (not SQLite's
shadow-table-recreate pattern) and don't depend on a PRAGMA
to disable referential integrity mid-migration.

Note: `TestSchemaEquivalence/sqlite` is currently broken at
HEAD for a separate, pre-existing reason traced to the
`cf28126` regen (the test fails both with and without this
fix). Will be investigated as a follow-up.

fix(cache): add retry-on-deadlock to storeNarInfoInDatabase

PR #1203 (PRRT_kwDONV6tZs6DsScG, medium): standalone
`storeNarInfoInDatabase` (Cache-free entry used by
MigrateNarInfo and the `ncps migrate-narinfo` CLI) called
`dbClient.WithTransaction` directly, skipping the
deadlock-retry loop the `*Cache.withEntTransaction` provides.
Under load — concurrent migrations of many narinfos against
SQLite, or contended PostgreSQL deployments — that meant
transient SQLITE_BUSY / serialization failures bubbled all the
way out, where the legacy path would have retried.

Fix: extract the retry policy from `*Cache.withEntTransaction`
into a package-level `withEntTransactionRetry(ctx, dbClient,
operation, fn)` helper. `*Cache.withEntTransaction` now
forwards to it; `storeNarInfoInDatabase` uses it directly.
Same retry parameters (5 attempts, 50ms initial back-off
doubling per attempt, ctx-aware via time.Timer + Stop).

No behaviour change for any caller already going through
`*Cache.withEntTransaction`; the previously-missing retry
is now present for MigrateNarInfo / CLI migrations.

fix(cache): batch HashIn + null-safe SUM + unshadow ent

PR #1204 surfaced three concerns in §11.4's non-tx NarFile
conversion (`feat-cache-narfile-non-tx-to-ent`):

1. (PRRT_kwDONV6tZs6DsRgI, high) The CDC delayed-cleanup
   replacement for `GetOldCompressedNarFiles` was a two-step
   query that paged ALL chunked NAR hashes into Go memory then
   passed them to a single `entnarfile.HashIn(...)`. On caches
   with many chunked NARs that risks (a) loading megabytes of
   hashes into memory and (b) blowing past driver parameter
   limits (Postgres ≤ 65535, modern SQLite ≤ 32766, older
   SQLite ≤ 999). Fixed by batching HashIn in chunks of
   `cdcCleanupHashBatchSize` (= 500, a new package const).
   The batch size is well below all known limits and well
   above typical chunked-hash counts, so the job stays a
   single round-trip for any practical install.

2. (PRRT_kwDONV6tZs6DsRgL, medium) The metrics-observer SUM
   aggregator scanned into a plain `int64`. SQL `SUM` over an
   empty set returns NULL, which fails to scan into `int64`
   ("sql: Scan, storing driver.Value type <nil> into type
   *int64"). The metric callback would have started failing
   the moment any deployment booted with an empty nar_files
   table. Switched to `sql.NullInt64` with a Valid check;
   NULL is treated as zero. The `sql:"sum"` tag matches the
   alias Ent emits via `ent.Sum`.

3. (PRRT_kwDONV6tZs6DsRgO, medium) Local var `ent` shadowed
   the `ent` package import inside the cron handler closure.
   No actual breakage in the current code (no `ent.X` use
   inside that scope) but it makes the scope brittle for
   future edits. Renamed to `entClient`.

`go test -race ./pkg/cache/...` passes against SQLite +
Postgres + MySQL + Redis. Lint clean.

fix(cache): batch relinkNarInfosToNarFile bulk insert

PR #1205 (PRRT_kwDONV6tZs6DsRuq, medium):
`relinkNarInfosToNarFileWithEntTx` passed every queried narinfo
ID into a single `tx.NarInfoNarFile.CreateBulk(...)` call. Each
row sends 2 parameters (narinfo_id, nar_file_id), so a closure
with N narinfos sharing a NAR URL sends 2N placeholders. On
older SQLite builds (limit 999) that fails for any N > 499;
even on Postgres (limit 65535) very large closures could
silently overflow.

Fixed by chunking the loop into batches of 500 rows = 1000
placeholders per batch, well below all driver limits and well
above realistic closure sizes — so the common case stays a
single round-trip.

batch size is a local const inside the helper rather than a
package constant because the limit derives from this specific
row's placeholder count (2 cols × batch size); other bulk
inserts with different column counts get their own limit.

`go test -race ./pkg/cache/...` passes against SQLite +
Postgres + MySQL + Redis. Lint clean.

### DIFF
--- a/migrations/sqlite/20260521021523_ent_baseline.sql
+++ b/migrations/sqlite/20260521021523_ent_baseline.sql
@@ -1,4 +1,10 @@
 -- +goose Up
+-- +goose NO TRANSACTION
+-- Required so the `PRAGMA foreign_keys = off` below takes effect —
+-- SQLite ignores the PRAGMA inside a transaction, which would cause
+-- the subsequent DROP TABLE / RENAME swaps to fail against any data
+-- already referencing the rewritten tables.
+
 -- disable the enforcement of foreign-keys constraints
 PRAGMA foreign_keys = OFF;
 -- create "new_narinfos" table
--- a/migrations/sqlite/atlas.sum
+++ b/migrations/sqlite/atlas.sum
@@ -1,4 +1,4 @@
-h1:ieTAmZaBCo20zXl2vHvJ5HYS58ZxtY83HtzgmAcJ0/w=
+h1:wd3qrlObmzyy0FSwtarrnKR0ujKdgdPexuDxTYHHHsc=
 20241210054814_create-narinfos-table.sql h1:e8MnIArqBCoUNv8/b0yDnx6ikbaSoPuMp3+j+C/cIPk=
 20241210054829_create-nars-table.sql h1:odrcFJuEF0MT6AIEa5Vn8ghpHV7EhIwfOjsIal1ZUW0=
 20241213014846_add-query-to-nars-table.sql h1:gFPvhup77Qua+8KlsWxqRLQqbXSr1IZSnpVDOFlR5cM=
@@ -13,4 +13,4 @@ h1:ieTAmZaBCo20zXl2vHvJ5HYS58ZxtY83HtzgmAcJ0/w=
 20260217071237_add_chunking_started_at_to_nar_files.sql h1:bUn8AnQH1Hc9burg/VqDqK5Yjs0JvjjK651p2+896aY=
 20260301000000_add_verified_at_to_nar_files.sql h1:NxNO9QJIibslyviRRNK3K2MkBAddVrXLsP2NSReaucc=
 20260318053903_add_pinned_closures_table.sql h1:9rWx0agGLMR6TDaq1jKPZNKlch4LOnKjb/OaHO5nPmI=
-20260521021523_ent_baseline.sql h1:ARI5SPrx9vpGf6cXv5nLf6Iccor3OIjTyWHrqShpNmQ=
+20260521021523_ent_baseline.sql h1:PkCUJB7uREMN8ERWLxSj3msPhjXZYTU3E6PUSmNOGBI=
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -4904,7 +4904,7 @@ func storeNarInfoInDatabase(
 	hash string,
 	narInfo *narinfo.NarInfo,
 ) error {
-	return dbClient.WithTransaction(ctx, "storeNarInfoInDatabase", func(tx *ent.Tx) error {
+	return withEntTransactionRetry(ctx, dbClient, "storeNarInfoInDatabase", func(tx *ent.Tx) error {
 		nir, err := upsertNarInfoFromParsed(ctx, tx, hash, narInfo)
 		if err != nil {
 			return err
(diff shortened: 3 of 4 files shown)


kalbasit:

- #1178
    - #1179
        - #1180
            - #1181
                - #1182
                    - #1183
                        - #1184
                            - #1185
                                - #1186
                                    - #1187
                                        - #1188
                                            - #1190
                                                - #1191
                                                    - #1192
                                                        - #1193
                                                            - #1194
                                                                - #1195
                                                                    - #1196
                                                                        - #1197
                                                                            - #1198
                                                                                - #1199
                                                                                    - #1200
                                                                                        - #1201
                                                                                            - #1202
                                                                                                - #1203
                                                                                                    - #1204
                                                                                                        - #1205
                                                                                                            - #1206
                                                                                                                - #1207
                                                                                                                    - #1208 ◀
                                                                                                                        - #1209
                                                                                                                            - #1210
                                                                                                                                - #1211
                                                                                                                                    - #1212
                                                                                                                                        - #1213
                                                                                                                                            - #1214
                                                                                                                                                - #1215
                                                                                                                                                    - #1216
                                                                                                                                                        - #1217
                                                                                                                                                            - #1218
                                                                                                                                                                - #1219
                                                                                                                                                                    - #1220
                                                                                                                                                                        - #1221
                                                                                                                                                                            - #1222
                                                                                                                                                                                - #1223
                                                                                                                                                                                    - #1224
                                                                                                                                                                                        - #1225
                                                                                                                                                                                            - #1226
                                                                                                                                                                                                - #1227
                                                                                                                                                                                                    - #1228
                                                                                                                                                                                                        - #1229
                                                                                                                                                                                                            - #1231
                                                                                                                                                                                                                - #1232
                                                                                                                                                                                                                    - #1233
                                                                                                                                                                                                                        - #1234

PR #1228, opened 2026-05-21
docs(openspec): archive migrate-to-ent-and-atlas

kalbasit:
Verify ran clean (109/109 tasks, valid). Five ent-lint
invariants without a triggering pattern in ncps today (A3, A5,
snake_case enum, expand-contract, CHECK-presence) were descoped
from this change and deferred to a future change that
introduces a matching schema pattern. The ent-schema-lint spec
and tasks.md were trimmed so the spec promise matches the
shipped implementation.

Archive synced delta specs to main: rewrote data-model (dropped
sqlc/dbmate requirements, swapped Querier -> *database.Client,
added Ent schemas as source of truth), created
database-migrations, database-orm, and ent-schema-lint
capabilities under openspec/specs/, then moved the change to
openspec/changes/archive/2026-05-21-migrate-to-ent-and-atlas/.

PR #1226, opened 2026-05-21
docs(readme,changelog): document Ent + Atlas + Goose; add operator upgrade notes

kalbasit:
Two doc-only changes:

- README.md
  - Replace the docker quick-start `dbmate up` invocation with
    `ncps migrate up` (the new migrator embeds the migrations and
    applies them via goose).
  - Add an Architecture section listing the new internal dependencies
    (Ent, Atlas-as-library, Goose, Chi, storage backends) and pointing
    at the docs for the full design.

- CHANGELOG.md (new file)
  - Document the database tooling migration (sqlc + dbmate → Ent +
    Atlas + Goose), the removal of `dbmate`/`dbmate-wrapper` and
    the `pkg/database/*db/` wrapper packages.
  - Operator upgrade procedure: backup first, then the one-shot
    adoption that converts the legacy `schema_migrations` table to
    the goose shape on first `ncps migrate up`. Calls out the
    SQLite/Postgres transactional path and the MySQL backup-table
    dance, plus the `--dry-run` preview for large databases.
  - Lists the four new `nix flake check` derivations.

Marks tasks 14.7 and 14.8 complete in the migrate-to-ent-and-atlas
change.

PR #1225, opened 2026-05-21
docs(skills): rewrite migration skills + add ent-schema skill

kalbasit:
Rewrites the .agent/skills/ tree to reflect the Ent + Atlas + Goose
workflow:

- `.agent/skills/migrate-new/SKILL.md` now drives the
  "edit ent/schema → go generate ./ent → cmd/generate-migrations"
  pipeline, with `--sql-only` for backfills + the four-step NOT NULL
  recipe, plus the `--skip` / `--postgres-url` / `--mysql-url`
  overrides for hand-translated migrations.
- `.agent/skills/migrate-up/SKILL.md` documents `ncps migrate up`
  with `--dry-run` preview, the one-shot dbmate adoption, and the
  three named error states (CorruptState, ImpossibleState, DownNotSupported).
- `.agent/skills/migrate-down/SKILL.md` becomes a doc on the
  expand-contract policy + four-step NOT NULL recipe; there is no
  `migrate down` command.
- Adds `.agent/skills/ent-schema/SKILL.md` documenting the five
  codegen invariants (A1-A5) with concrete examples, cross-referencing
  cmd/ent-lint for live AST enforcement of A1, A2, A4.
- Deletes `.agent/skills/sqlc/`, `.agent/skills/generate-db-wrappers/`,
  and `.agent/skills/dbmate/`.

Marks tasks 14.2-14.6 complete in the migrate-to-ent-and-atlas change.

PR #1224, opened 2026-05-21
docs(claude): rewrite CLAUDE.md for the Ent + Atlas + Goose workflow

kalbasit:
Replaces every reference to the removed sqlc / dbmate /
dbmate-wrapper / db/query.*.sql / db/schema/ / db/migrations/ tooling
with the Ent + Atlas + Goose workflow:

- New "Common Commands" section covers `go generate ./ent/...`,
  `go run ./cmd/ent-lint`, `task ent:check`,
  `go run ./cmd/generate-migrations` (schema-driven + --sql-only),
  `ncps migrate up` (with --dry-run), and `go run ./cmd/atlas-sum-check`.
- "Skills" block lists the new ent-schema / migrate-new / migrate-up /
  migrate-down skills.
- "Package Structure" reorganized around `cmd/{ent-lint,generate-migrations,atlas-sum-check}`,
  `ent/schema/` (hand-authored) vs `ent/` (committed codegen), and
  `migrations/<dialect>/` with embedded `atlas.sum` integrity files.
- New "Database" section documents the expand-contract policy (column
  changes split across N migrations), the four-step NOT NULL recipe,
  and the five Ent codegen invariants (A1-A5) enforced by
  `cmd/ent-lint`.

Marks task 14.1 complete in the migrate-to-ent-and-atlas change.

PR #1229, opened 2026-05-21
refactor(cache): scope stale-lock orphan query to known candidate IDs

kalbasit:
refactor(cache): scope stale-lock orphan query to known candidate IDs

Restrict the orphan-check query inside cleanupStaleLockChunks to the chunk
IDs we already know about from staleLockChunks, instead of pulling every
orphan in the database and filtering them in memory. The post-filter
`if _, ok := staleByID[oc.ID]; !ok` is no longer needed since the query
already returns only candidates.

Addresses gemini-code-assist comment on PR #1209 (PRRT_kwDONV6tZs6D3VF-).

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

refactor(fsck): drop redundant seen map in CDC size-mismatch check

queryCDCNarFilesWithSizeMismatch's `seen` map was redundant: Ent's
All(ctx) returns one row per NarFile, and the inner loop already breaks
after recording the first mismatch, so each NarFile can only be appended
to mismatched once.

Addresses gemini-code-assist comment on PR #1212 (PRRT_kwDONV6tZs6D3P8r).

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

test(database): use t.Context() in freshSchemaSQLite helper

Replace context.Background() with t.Context() so the schema-create call
is bound to the test lifetime. Drops the now-unused context import.

Addresses gemini-code-assist comment from PR #1214
(PRRT_kwDONV6tZs6D3PyX); the original test_schema_parity file was folded
into client_test.go but the same pattern lives in freshSchemaSQLite.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

refactor(cache): explicit not-found vs error split in migration double-check

After acquiring the migration lock, distinguish ent.IsNotFound (the
expected "not yet migrated" case, fall through to migrate) from other DB
errors (log at warn level instead of silently swallowing). The migration
still attempts on any error so a transient lookup glitch doesn't block
progress; if the DB is genuinely broken the upsert below will surface
the failure.

Addresses gemini-code-assist comment on PR #1216 (PRRT_kwDONV6tZs6D3YoD).

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

refactor(analytics): scan SUM(file_size) into a single sql.NullInt64

Ent's Aggregate(ent.Sum(...)) emits a raw SUM(...) expression without an
alias; scanning into []struct{ Sum sql.NullInt64 \`sql:"sum"\` } relied
on tag-name resolution that may not match the column header. Scan
positionally into a single sql.NullInt64 instead, which is more robust
and also correctly handles the NULL result that SUM returns on an empty
table.

Addresses gemini-code-assist comments on PR #1217
(PRRT_kwDONV6tZs6D3Um1, PRRT_kwDONV6tZs6D3Um6).

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

docs(database): correct IsNotFoundError comment about delegation

The previous wording referenced a "well-known Ent sentinel below" that
doesn't exist in this file — IsNotFoundError simply delegates to
isEntNotFound (declared in ent_errors.go) which keeps the ent import
isolated to a single file.

Addresses gemini-code-assist comment on PR #1218 (PRRT_kwDONV6tZs6D3WsX).

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

build(formatter): re-include ent/generate.go in the formatter glob

ent/generate.go is hand-written (carries the //go:generate directive)
but the broad ent/*.go exclusion swept it into the codegen-only group.
Add an explicit !ent/generate.go negation so it stays formattable
without re-introducing diff loops against the codegen drift check.

Addresses gemini-code-assist comment on PR #1221 (PRRT_kwDONV6tZs6D3zlq).

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

docs(atlas-sum-check): correct doc comment about exit behavior

The package doc claimed the binary exits non-zero on the first mismatch.
The implementation actually iterates every dialect and only exits at the
end, reporting all offending directories at once. Update the doc to
match.

Addresses gemini-code-assist comment on PR #1222 (PRRT_kwDONV6tZs6D31Je).

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

build(checks): quote \$src paths in schema-equivalence preCheck

Quote the path expansions when sourcing the pre/post-check helper
scripts in schema-equivalence-check. Nix store paths don't contain
spaces in practice, but quoting is shell best practice and silences
shellcheck-style nitpicks on the inherited shell content.

Addresses gemini-code-assist comment on PR #1223 (PRRT_kwDONV6tZs6D32tm).

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

docs(claude): use host:port/database in URL examples + clarify migrate down

- Bring the Common Commands DSN examples in line with the Database section
  by using the full host:port/database placeholder for postgresql/mysql.
- Replace "there is no migrate down command" with the accurate phrasing:
  the command is implemented (pkg/ncps/migrate.go:107) but returns an
  error pointing at the expand-contract policy.

Addresses gemini-code-assist comments on PR #1224 (PRRT_kwDONV6tZs6D37b-,
PRRT_kwDONV6tZs6D37cD). The third sqlfluff suggestion is intentionally
skipped — migrations/ is sealed by atlas.sum so sqlfluff isn't run
against it, and the only SQL outside migrations/ is already covered by
nix fmt.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

docs(ent-schema): A2 is the OnDelete invariant, not enum naming

Both .agent/skills/ent-schema/SKILL.md and CLAUDE.md described A2 as the
snake_case enum-type convention, which contradicted cmd/ent-lint/main.go
(line 286+) and openspec/specs/ent-schema-lint/spec.md — both of which
treat A2 as the "OnDelete on edge.From is forbidden" check.

Reword A2 in both docs to match the actual linter invariant, and move
the snake_case enum-type rule into its own section labelled as a
convention enforced by code review until the schema tree gains its
first field.Enum(...) declaration.

Addresses gemini-code-assist comment on PR #1225 (PRRT_kwDONV6tZs6D34eU).

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

docs(changelog): mention task migrations:gen and clarify row-count check

- Lead with the Taskfile entrypoint (task migrations:gen NAME=...) for
  generating new migrations, since it regenerates the Ent client first
  via task ent:generate. Mention the underlying go run invocation in
  parentheses for users who want the raw command.
- The schema_migrations adoption verifies row-count consistency, not
  parity — the new table intentionally adds a sentinel row, so the
  expected post-count is preCount + 1. Reword to make that explicit.

Addresses gemini-code-assist comments on PR #1226
(PRRT_kwDONV6tZs6D3756, PRRT_kwDONV6tZs6D376E).

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

docs(openspec): align migration + lint specs with actual implementation

Four wording corrections so the OpenSpec specs match what the code does:

- Fresh-install seeding text claimed the Go side passes a time.Now()
  to schema_migrations.tstamp. Goose's database.InsertRequest exposes
  only Version, so the tstamp column relies on the database's default
  timestamp function. Reword the Requirement and Scenario lines to
  match.
- The "Forbidden DDL" scenario claimed cmd/ent-lint SHALL fail on a
  forbidden DDL statement. The linter has not yet been wired up for
  that check (see cmd/ent-lint/main.go:21 deferral comment and the
  scope note already documented in ent-schema-lint/spec.md). Update
  the THEN clause to describe the current manual-review enforcement,
  retaining the future-cmd/ent-lint intent.
- The example output in ent-schema-lint/spec.md no longer matches
  cmd/ent-lint/main.go:511-515 — the real output includes the source
  line number and a fuller "Ent will fabricate a phantom FK column"
  trailer. Update the example accordingly.

Addresses gemini-code-assist comments on PR #1228
(PRRT_kwDONV6tZs6D4drm, PRRT_kwDONV6tZs6D4dr2, PRRT_kwDONV6tZs6D4dsA,
PRRT_kwDONV6tZs6D4dsG).

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/.agent/skills/ent-schema/SKILL.md
+++ b/.agent/skills/ent-schema/SKILL.md
@@ -32,16 +32,30 @@ func (Chunk) Annotations() []schema.Annotation {
 }
 ```
 
-### A2 — snake_case enum-type naming
+### A2 — `entsql.OnDelete(...)` lives on `edge.To`, never on `edge.From`
 
-Ent uses snake_case enum-type names by convention. Every `field.Enum(...)`
-needs a matching `entsql.Annotation{Type: "<table>_<column>_enum"}` so
-the generated DDL emits a stable, predictable type name across dialects.
+Ent silently ignores `entsql.OnDelete(...)` annotations attached to
+`edge.From(...)` declarations. CASCADE / RESTRICT / SET NULL semantics
+MUST be declared on the owning `edge.To(...)` side, otherwise the
+generated DDL falls back to the default ON DELETE rule and the intended
+behaviour disappears without warning.
 
 ```go
-field.Enum("compression").
-    Values("none", "xz", "zstd").
-    Annotations(entsql.Annotation{Type: "nar_files_compression_enum"})
+// Correct — OnDelete on edge.To (the FK owner).
+func (Parent) Edges() []ent.Edge {
+    return []ent.Edge{
+        edge.To("children", Child.Type).
+            Annotations(entsql.Annotation{OnDelete: entsql.Cascade}),
+    }
+}
+
+// Wrong — OnDelete on edge.From is silently dropped.
+func (Child) Edges() []ent.Edge {
+    return []ent.Edge{
+        edge.From("parent", Parent.Type).Ref("children").Unique().
+            Annotations(entsql.Annotation{OnDelete: entsql.Cascade}),
+    }
+}
 ```
 
 ### A3 — UNIQUE + edge.From().Field() pairing
@@ -81,6 +95,20 @@ func (Child) Edges() []ent.Edge {
 Every `field.Bytes("*_ciphertext")` field MUST chain `.Sensitive()` so
 the ciphertext never leaks into error messages, log lines, or the
 generated `String()` methods.
+
+### Snake_case enum-type naming (convention; not yet linted)
+
+Ent uses snake_case enum-type names by convention. Every `field.Enum(...)`
+needs a matching `entsql.Annotation{Type: "<table>_<column>_enum"}` so
+the generated DDL emits a stable, predictable type name across dialects.
+This is enforced by code review until ncps's schema tree gains its first
+`field.Enum(...)` declaration and the check is wired into `cmd/ent-lint`.
+
+```go
+field.Enum("compression").
+    Values("none", "xz", "zstd").
+    Annotations(entsql.Annotation{Type: "nar_files_compression_enum"})
+```
 
 ```go
 field.Bytes("password_ciphertext").Sensitive()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,11 @@ project loosely follows [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - **Database tooling migrated from sqlc + dbmate to Ent + Atlas + Goose.**
   Schemas are now authored under `ent/schema/*.go`, migrations are
   generated from Atlas diffs (used as a Go library) via
-  `go run ./cmd/generate-migrations --name=<descriptive_snake_case>`,
-  and applied at runtime by `ncps migrate up`. The runtime applier is
-  `goose.NewProvider` against the embedded `migrations/<dialect>/` FS.
+  `task migrations:gen NAME=<descriptive_snake_case>` (which regenerates
+  the Ent client via its dependency on `ent:generate`, then calls
+  `go run ./cmd/generate-migrations --name=...`), and applied at runtime
+  by `ncps migrate up`. The runtime applier is `goose.NewProvider`
+  against the embedded `migrations/<dialect>/` FS.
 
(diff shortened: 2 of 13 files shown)


kalbasit:

- #1226
    - #1227
        - #1228
            - #1229 ◀
                - #1231
                    - #1232
                        - #1233
                            - #1234

kalbasit:
Automated stack merge failed during the rebase step. The `merge-stack` label has been removed from all PRs in the stack. Please resolve conflicts locally, push, and re-apply the label on the top-most PR to retry. Run: https://github.com/kalbasit/ncps/actions/runs/26326665317

PR #1231, opened 2026-05-22
chore(nix): restore dbmate to the dev shell only

kalbasit:
dev-scripts/run.py's perform_clean() uses `dbmate drop` to reset
dev databases between scenarios. dbmate was removed in 80c4a15
along with sqlc; restore it to the dev shell only — keep it out
of nix/dev-packages.nix (shared with the docker-dev image) and
nix/packages/ncps/default.nix. Runtime migration is still
`ncps migrate up` (embedded goose).

PR #1227, opened 2026-05-21
chore(openspec): mark §14.9 final fmt + lint sweep complete

kalbasit:
Final pass verified: `nix fmt && golangci-lint run --fix ./...` are
both no-ops at this point — the per-commit hooks ran throughout the
§13/§14 work and the tree is clean.

Full test suite (`go test -race -count=1 -timeout 12m ./...`) passes
across SQLite, PostgreSQL, MySQL, MinIO/S3, and Redis with all
integration test gates enabled.

Marks task 14.9 complete; the migrate-to-ent-and-atlas change is now
done modulo the deferred §5 ent-lint invariants tracked elsewhere.

PR #1232, opened 2026-05-22
fix(dev-scripts): use `ncps migrate up` in run.py

kalbasit:
The current branch's migrations live in migrations/<dialect>/ in
goose format; run.py's `dbmate up` step was silently no-opping
because dbmate's default migrations dir is db/migrations/, which
no longer exists. Switch to `go run . migrate up` so the script
exercises the real embedded migration runner — and, against an
existing dbmate-shaped `schema_migrations` table, the dbmate→
goose adoption path in pkg/database/migrate/.

For non-sqlite backends, add a `dbmate create` step before the
migration. `dbmate up` used to auto-create the database; `ncps
migrate up` only connects to an existing one. `dbmate create`
is no-op on first failure (database already exists) and errors
are suppressed, so the call is idempotent in practice. SQLite is
handled by ensuring the parent directory exists; the goose runner
creates the file on first connect.

### DIFF
--- a/dev-scripts/run.py
+++ b/dev-scripts/run.py
@@ -359,12 +359,33 @@ def check_dependencies(args):
 
 def run_db_migration(db_url):
     log(f"Migrating database: {db_url}", YELLOW)
-    # Ensure directory exists for sqlite
+    # Ensure directory exists for sqlite.
+    # For postgres/mysql, ensure the database itself exists — `dbmate
+    # up` used to do this implicitly, but `ncps migrate up` connects
+    # to an existing database and won't create it. We use `dbmate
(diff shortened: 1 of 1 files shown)


kalbasit:

- #1226
    - #1227
        - #1228
            - #1229
                - #1231
                    - #1232 ◀
                        - #1233
                            - #1234

kalbasit:
Automated stack merge failed during the rebase step. The `merge-stack` label has been removed from all PRs in the stack. Please resolve conflicts locally, push, and re-apply the label on the top-most PR to retry. Run: https://github.com/kalbasit/ncps/actions/runs/26326665317

PR #1235, opened 2026-05-22
chore: migrate CI to kalbasit/gh-actions reusable workflow

kalbasit:
Replace per-repo CI plumbing (filter, generate, build, multi-arch
manifest, codecov, final gate) with a thin caller of the shared
reusable workflow at
kalbasit/gh-actions/.github/workflows/ci.yml@feat-add-reusable-ci-workflows.
Delete the per-repo build.yml; the shared workflow handles
flake-check, coverage, codecov upload, multi-arch OCI image build/
push, and manifest creation.

ncps-specific work (sqlc database codegen and Cloudflare Pages docs
deploy) stays in this file as sibling jobs because it doesn't
generalize across the other consumers. The sibling jobs run their
own dorny/paths-filter since the reusable workflow doesn't (today)
re-expose filter outputs to the caller.

A local final-gate job wraps both the shared workflow and the
sibling jobs so branch protection only needs to require one check.

Net diff: -245 / +68. No behavior change expected; will repin to
@v0.1.0 once kalbasit/gh-actions tags it.


kalbasit:

- #1235 ◀

PR #1234, opened 2026-05-22
fix(helm,docs): use `ncps migrate up`, drop dbmate

kalbasit:
The dbmate→Ent migration removed dbmate from the runtime container
image (only /bin/ncps ships now), but the Helm chart's three migration
entrypoints and the _helpers.tpl partial still invoked /bin/dbmate
with DATABASE_URL. Helm deploys of the post-Ent branch would crash
with `exec: "/bin/dbmate": no such file or directory`. Several
developer + user docs also described the dbmate-era workflow and
db/migrations/<dialect> paths that no longer exist.

Changes:

* charts/ncps/templates/_helpers.tpl: ncps.migrationDatabaseURLEnv
  now emits CACHE_DATABASE_URL (the env-var source already declared
  by the --cache-database-url flag in pkg/ncps/migrate.go). Secret
  key name (database-url) is unchanged.
* charts/ncps/templates/{migration-job,statefulset,deployment}.yaml:
  command "/bin/ncps", args ["migrate","up"].
* charts/ncps/tests/migration_test.yaml (new): 6 helm-unittest cases
  asserting command/args/env for migration-job + both initContainer
  forms, covering sqlite (literal value) and postgres (secretKeyRef).
  Locks down the new spec contract added in this change.
* docs: replace every dbmate invocation across 12 files (developer +
  user guides, Docker / Docker Compose / Kubernetes / NixOS install
  recipes, troubleshooting) with the equivalent
  `ncps migrate up --cache-database-url=...` command. Updated the
  Contributing.md project tree to reflect ent/, migrations/, and the
  cmd/{ent-lint,generate-migrations,atlas-sum-check} helpers. Added
  an "Upgrading past the Ent migration release" subsection to
  Operations/Upgrading.md.
* openspec/changes/update-migration-docs-and-chart/: full
  spec-driven artifact set (proposal, design, specs delta on
  database-migrations, tasks). The specs delta adds a new requirement
  pinning the runtime migration entrypoint to /bin/ncps so any
  future deployment artifact knows to avoid dbmate.

Verified: nix fmt clean, helm unittest charts/ncps green (135/135),
no dbmate references remain in charts/ or docs/ outside the
intentional upgrade-note callout.

PR #1233, opened 2026-05-22
test(dev-scripts): add e2e migration test orchestrator

kalbasit:
dev-scripts/test-migration-e2e.py drives a main→current-branch
migration test across 12 scenarios (3 DBs × 2 CDC modes × 2 storage
backends). For each scenario it:

  1. Resets var/ncps + drops dev DBs + recreates the S3 bucket.
  2. Detaches HEAD at main, starts ncps via dev-scripts/run.py.
  3. Seeds the cache via dev-scripts/nix-isolated-build.py with two
     small Nix packages.
  4. Snapshots tables + narinfo hashes + pubkey.
  5. Switches back to the migration branch and runs
     `go run . migrate up --dry-run` then `migrate up` then
     `--dry-run` again, asserting idempotency.
  6. Snapshots again and verifies every pre-migration narinfo hash
     is still present.
  7. Restarts ncps on the migration branch and verifies the cache
     still serves a pre-existing entry (cache-hit) AND can accept a
     new write (write-path).

The detached-HEAD checkout avoids conflicting with the main worktree
in this repo. Per-scenario nix temp stores are routed onto the real
disk via TMPDIR=var/ncps/nix-tmp so a tmpfs /tmp can't exhaust the
test run after a handful of scenarios. Result logs land under
.e2e-results/<timestamp>/ which is now in .gitignore.

PR #1241, opened 2026-05-23
test: shrink go test wall time ~48%

kalbasit:
`nix flake check` and `go test -race ./...` were slow enough (~88s
on a quiet machine) to discourage running the full suite locally.
This change keeps every assertion but stops paying for real network
sleeps, SDK retry backoffs, oversized payloads, and per-test go
builds.

Profiling tool:
- Add dev-scripts/profile-tests.py to rank packages and tests by
  wall time from `go test -json` output. Kept as a permanent project
  tool so future drift is easy to spot.

pkg/cache/upstream (13.6s -> 2.0s):
- The four "timeout if server takes more than 3 seconds before first
  byte" subtests now pass an explicit ResponseHeaderTimeout=50ms and
  stall the server only 500ms, instead of waiting on the production
  default 3s timeout.
- The "custom DialerTimeout" subtest previously used a slow-Accept
  listener, which does NOT delay the dialer (the kernel auto-
  completes the SYN handshake), so the original test was indirectly
  hitting ResponseHeaderTimeout. Rewrite it to dial 192.0.2.1
  (TEST-NET-1, RFC 5737) with DialerTimeout=50ms, which actually
  trips the dialer's timer.

pkg/storage/s3 (12.2s -> 1.5s):
- minio-go retries RoundTripper errors with exponential backoff,
  taking 3-5s per ErrorPath test (15+ tests).
- Replace those mocks with a new s3ErrorResponse() helper that
  returns a non-retryable 403 AccessDenied. Where assertions check
  for sentinel text ("put failed", "list objects failed", etc.) the
  message is embedded in the response body so minio-go surfaces it
  verbatim into err.Error() and existing assertions keep working.

pkg/cache (16.5s -> ~4s):
- DeadlockContextCancellationDuringDownload: 1MB / 1024-chunk /
  10ms-per-chunk payload shrunk to 50KB / 50-chunk / 5ms-per-chunk.
  Background-completion timeouts trimmed from 10s/15s to 3s/3s.
- Four 1s sleeps in last-accessed-at tests cut to 100ms; all
  supported engines store timestamps with sub-second precision.
- TestIssue990 upstream sleep 1s -> 200ms, post-cancel wait 1.5s ->
  500ms.
- TestCDCBackends first-pull chunking delay 2s -> 500ms.

pkg/lock/redis (7.8s -> 2.0s):
- TestNewLocker_ReturnType wrapped each NewLocker call in a 500ms
  context so the degraded-mode Ping retries don't burn the go-redis
  default DialTimeout (5s) per attempt.

cmd/generate-migrations (5.1s -> 3.0s):
- buildGenerateMigrations now uses sync.Once + TestMain so the
  binary is built once per package run instead of per-test.

Also scaffolds the OpenSpec change (proposal, design, two specs,
tasks) that drove this work, plus the baseline / post-batch timing
captures used to verify progress.

All edits keep the race detector on; lint is clean; every
re-profiled package passed \`go test -race\`.

### DIFF
--- a/cmd/generate-migrations/main_test.go
+++ b/cmd/generate-migrations/main_test.go
@@ -1,12 +1,27 @@
 package main_test
 
 import (
+	"context"
+	"os"
 	"os/exec"
 	"path/filepath"
+	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+)
+
+// sharedBinary is built once for the whole package via TestMain so each test
+// reuses the same executable. Building it per-test cost ~3-4s × N.
+//
+//nolint:gochecknoglobals // intentional package-level state for one-shot binary build
+var (
+	sharedBinary    string
+	errSharedBuild  error
+	sharedBuildOnce sync.Once
+	sharedBuildOut  string
+	sharedBinaryDir string
 )
 
 // TestSQLOnlyEmitsThreeDialects verifies the --sql-only path produces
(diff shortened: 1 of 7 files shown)


kalbasit:

- #1231
    - #1232
        - #1233
            - #1234
                - #1241 ◀
                    - #1242
                        - #1243
                            - #1244
                                - #1245
                                    - #1246

PR #1242, opened 2026-05-23
test: shrink fsck/security/large-narinfo tests

kalbasit:
Continues the previous commit's work. Each remaining outlier had a
concrete cause; this round drops the full suite from ~45.5s to
~41.6s, putting it past the stretch >=50% reduction target vs the
original 88.1s baseline.

pkg/ncps/fsck_test.go TestFsckBackends/SQLite/VerifiedSince
(2.05s -> 0.15s):
- The "wait long enough that --verified-since 1s actually trips"
  step relied on a 2-second sleep so that the subsequent fsck call
  with --verified-since=1s would not skip. Sub-second timestamp
  precision is available on every supported engine, so swap to a
  100ms sleep and --verified-since=1ms.

pkg/ncps/migrate_narinfo_test.go LargeNarInfo (1.31s -> 0.01s):
- The "large" narinfo had 100 references, 50 signatures, and a 1 MB
  random NAR body. Migration only walks narinfo files and inserts
  refs/sigs; the NAR body is never read. Shrink references to 20,
  signatures to 10, NAR body to 1 KB; update spot-check indices to
  match. 20 refs is still far above the typical narinfo (2-5).

pkg/server/security_test.go TestSecurity (2.02s -> 0.42s):
- Four "should NOT reach upstream" subtests each waited a full 500ms
  to confirm the upstream channel stayed empty. An in-process
  httptest.Server roundtrip takes ~1ms, so 100ms is plenty of
  margin. Both arms (positive and negative) now use the same 100ms
  cap.

pkg/cache/cache_internal_test.go (last_accessed_at sleeps):
- Two more occurrences of "ensure time has moved by one sec" in the
  RunLRU / RunLRUCleanupInconsistentNarInfoState code paths cut from
  1s to 100ms, matching the change already made in cache_test.go in
  the previous commit.

pkg/cache/cache_internal_test.go testConcurrentDecompression
(1.13s -> ~0.3s):
- The 1 MiB random NAR + zstd compression pre-step dominated wall
  time; the concurrent-decompression assertion is just as strong
  with a 100 KiB payload and saves the bulk of the randomness +
  zstd cost.

Also captures the post-batch profiler output under
openspec/changes/less-tests/after-batch3-timings.txt.

Skipped intentionally:
- TestCDCChunker_Chunk_ErrorRace (10000 iterations to catch a
  historical race): already short-mode gated; reducing iterations
  would weaken the safety net.
- TestGetNarInfoDistributedCoordination (1s upstream sleeps): the
  test models inter-instance lock coordination and the 1s/2s
  timeouts give margin against CI load. Risk of flake outweighs the
  ~1s saving.

Verified with `go test -race -count=1` on every touched package;
\`golangci-lint run\` is clean; \`nix fmt\` shows no changes.

### DIFF
--- a/pkg/cache/cache_internal_test.go
+++ b/pkg/cache/cache_internal_test.go
@@ -464,8 +464,9 @@ func testRunLRU(factory cacheFactory) func(*testing.T) {
 			assert.True(t, found, nu.String()+" should exist in the store")
 		}
 
-		// ensure time has moved by one sec for the last_accessed_at work
-		time.Sleep(time.Second)
+		// Ensure time has advanced past timestamp precision for the last_accessed_at
+		// work. Every supported engine stores sub-second precision, so 100ms is plenty.
+		time.Sleep(100 * time.Millisecond)
 
 		// pull the nars except for the last entry to get their last_accessed_at updated
 		sizePulled = 0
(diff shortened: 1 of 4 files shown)


kalbasit:

- #1231
    - #1232
        - #1233
            - #1234
                - #1241
                    - #1242 ◀
                        - #1243
                            - #1244
                                - #1245
                                    - #1246

PR #1243, opened 2026-05-23
fix(ncps): close dbClient on shutdown; use FORCE drop in tests

kalbasit:
Found while shrinking the integration-test wall time. Every ncps
subcommand (serve, fsck, migrate-narinfo, migrate-nar-to-chunks)
opens a *database.Client via createDatabaseClient but never registers
its Close in the cli After-hook. In production the process just
exits so it's invisible; in tests the connection pool lingers, and
every integration-test cleanup paid a flat ~5s waiting for
DROP DATABASE to drain the leaked connections.

Quantified impact (full integration run, all deps live):
- pkg/ncps total: 23.07s -> 6.28s
- 18 TestFsckBackends/PostgreSQL/CDC/* tests all dropped off the
  "elapsed > 500ms" list (was uniformly ~5s each).
- Cumulative `go test -race ./...` with integration backends:
  80.09s -> 64.40s (~20% reduction in this commit alone).

Changes:
- pkg/ncps/{fsck,serve,migrate_narinfo,migrate_nar_to_chunks}.go:
  register `dbClient.Close` as a shutdown function right after
  createDatabaseClient. Captured by closure so the body of the
  Action continues to use the same client; After-hook calls Close
  in parallel with otelShutdown via wg.Go.
- nix/process-compose/postgres-dblink-create-drop-functions.sql:
  drop_test_db now issues `DROP DATABASE ... WITH (FORCE)` (PG 13+)
  as a belt-and-braces measure in case any future test still leaks
  a connection past Close. The Go-side fix above is the actual
  reason the per-test wait disappeared.

Also captures the integration timings before and after this commit
under openspec/changes/less-tests/integration-*-timings.txt.

Verified: `go test -race` on every touched package passes;
\`golangci-lint run\` is clean. The applied SQL function update was
also run against the live PG instance by hand so subsequent local
runs immediately benefit.

### DIFF
--- a/nix/process-compose/postgres-dblink-create-drop-functions.sql
+++ b/nix/process-compose/postgres-dblink-create-drop-functions.sql
@@ -34,13 +34,17 @@ END;
 $$ LANGUAGE plpgsql SECURITY DEFINER;
 
 -- 4. Function to DROP database
+-- Uses WITH (FORCE) (PostgreSQL 13+) so leaked connections from the test process
+-- (e.g. the inner dbClient pool used by ncps subcommands) don't make DROP DATABASE
+-- block for the connection idle timeout. Without FORCE, every per-test cleanup paid
+-- a ~5s wait for connections to drain.
 CREATE OR REPLACE FUNCTION drop_test_db(dbname text) RETURNS void AS $$
 BEGIN
     IF left(dbname, 5) != 'test-' THEN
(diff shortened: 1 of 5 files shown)


kalbasit:

- #1231
    - #1232
        - #1233
            - #1234
                - #1241
                    - #1242
                        - #1243 ◀
                            - #1244
                                - #1245
                                    - #1246

PR #1244, opened 2026-05-23
test: shrink integration outliers (s3 bucket, redis TTL, NAR)

kalbasit:
Continuation of the integration-test wall-time work. Builds on the
fix(ncps): close dbClient on shutdown commit, which already removed
~16s of DROP DATABASE wait. This commit picks off the remaining
named integration outliers:

pkg/storage/s3 TestNew_BucketAccessError (5.22s -> ~0s):
- The RoundTripper mock returned a Go error, triggering minio-go's
  exponential-backoff retry loop (~5s before giving up). Replace
  with s3ErrorResponse() so the SDK gives up immediately on the
  non-retryable 403. The production wrapper at s3.go:95 already
  wraps any non-nil err with "error testing bucket access", so the
  Contains() assertion still passes. The now-unused
  errTestingBucketAccess sentinel is removed.

pkg/cache TestDistributedBackends/.../LockFailover (4.51s -> ~1s
each, x3 backends):
- shortTTL: 2s -> 500ms (sleep matched at 800ms).
- ctx2 timeout: 500ms -> 200ms — still well under shortTTL so the
  "second locker fails while first holds the lock" check is sound.
- Total saving across the three backend variants: ~10s of test
  wall time.

pkg/cache TestDistributedBackends/.../LargeNARConcurrentDownload
(5.4s -> ~1s each, x3 backends):
- NAR payload 2 MiB -> 256 KiB. The test exercises concurrent
  download coordination, not raw throughput; 256 KiB still trips
  CDC chunking and keeps the lock dance honest.
- Server-side artificial delay 2s -> 300ms. The only requirement is
  that follower goroutines arrive while the first download is in
  flight, and goroutine scheduling latency is sub-millisecond on a
  quiet machine.

pkg/lock/redis TestLocker_LockExpiry (2.01s -> ~0.6s):
- Lock TTL 1s -> 200ms, post-expiry sleep 2s -> 400ms. Redis lock
  expiry is millisecond-granular; 400ms is plenty of safety.

pkg/lock/redis TestLocker_DegradedModeDisabled (1.72s -> ~0.5s):
- The "unreachable Redis port" call previously waited on the
  go-redis default DialTimeout (5s) per attempt. Wrap ctx in a
  500ms WithTimeout so the failure surfaces fast — matches the
  same fix already applied to TestNewLocker_ReturnType.

Skipped intentionally:
- TestDistributedBackends/.../PutNarInfoConcurrentSharedNar (50
  iterations to catch a concurrency race) — same policy as the
  chunker race test; reducing iterations weakens the safety net.

Cumulative integration-test wall time (full suite, all deps live):
~64.40s -> ~58.99s after this commit. Combined with the previous
dbClient fix, integration baseline 80.09s -> 58.99s (~26% lower).

Captures the post-batch profile under
openspec/changes/less-tests/integration-after-batch5-timings.txt.

Verified: \`go test -race\` on every touched package passes;
\`golangci-lint run\` is clean.

### DIFF
--- a/pkg/cache/cache_distributed_test.go
+++ b/pkg/cache/cache_distributed_test.go
@@ -471,7 +471,9 @@ func testDistributedLockFailover(_ distributedDBFactory) func(*testing.T) {
 		require.NoError(t, err)
 
 		testKey := "test-failover-key"
-		shortTTL := 2 * time.Second
+		// Short TTL keeps the "wait for expiry" step under a second; previously this was
+		// 2s + 2s safety = 4s of pure waiting per test, ×3 backends.
+		shortTTL := 500 * time.Millisecond
 
 		// Locker 1 acquires the lock
 		err = locker1.Lock(ctx, testKey, shortTTL)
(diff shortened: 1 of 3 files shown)


kalbasit:

- #1231
    - #1232
        - #1233
            - #1234
                - #1241
                    - #1242
                        - #1243
                            - #1244 ◀
                                - #1245
                                    - #1246

PR #1246, opened 2026-05-23
docs: fix awk/jq→Python and Nix sandbox spec

kalbasit:
Address two review comments from PR #1245:

- design.md: The profiling tool was implemented as Python
  (dev-scripts/profile-tests.py), not an awk/jq pipeline.
  Update the D1 rationale to reflect the actual implementation.

- spec.md: The "writes the ranking to a file" step contradicts
  running inside nix flake check, where the source tree is
  read-only (Nix sandbox). Qualify the requirement as local
  execution only so the spec no longer contradicts itself.

PR #1245, opened 2026-05-23
chore(openspec): archive less-tests change

kalbasit:
Adds the archived openspec change directory and updates the
short-test-mode and test-suite-efficiency specs to reflect the
shipped state of the stack.

PR #1223, opened 2026-05-21
feat(nix-checks): add schema-equivalence-check + refresh ci.yml

kalbasit:
Adds the `schema-equivalence-check` derivation that runs
`TestSchemaEquivalence` across SQLite, Postgres, and MySQL. Postgres
and MySQL are spun up via the existing `pre-check-{postgres,mysql}.sh`
helpers; cleanup is wired through a trap to `post-check-*.sh`. The
parent ncps `buildPhase` is replaced with a no-op so the check skips
the full binary build (only the migrations test pass is needed).

Drive-by: refresh `.github/workflows/ci.yml` to drop references to the
removed sqlc/dbmate tooling (`db/migrations/**`, `db/query.*.sql`,
`sqlc.yml`, `go generate ./pkg/database`) and replace them with the
Ent generator flow (`ent/schema/**`, `ent/generate.go`,
`go generate ./ent/...`). The build job is unchanged — it still runs
`nix flake check`, which now exercises all four new derivations.

Marks tasks 8.7, 13.4, 13.5, and 13.6 complete in the
migrate-to-ent-and-atlas change.

PR #1221, opened 2026-05-21
feat(nix-checks): add ent-codegen-drift-check + ent-lint-check

kalbasit:
Adds two new derivations to the nix checks tree, both of which feed into
`nix flake check`:

- `ent-codegen-drift-check` materializes the source tree, runs
  `go generate ./ent/...`, and fails the build if any committed file
  under `./ent` differs after regeneration. Uses `proxyVendor = true`
  because Ents `tool` dependency (entgo.io/ent/cmd/ent) is not vendored.
- `ent-lint-check` builds and runs `cmd/ent-lint --root .` and fails
  if any `[FAIL]` lines appear in the output.

Drive-by fixes:

- Update vendor hash to match the post-cleanup go.sum (pre-existing
  drift after the sqlc removal).
- Drop the now-broken `cp -r db $out/share/ncps/db` from postInstall;
  the `db/` tree no longer exists.
- Add `ent/*.go`, `ent/<generated-pkg>/**/*.go`, and
  `migrations/**/*.sql` to treefmt-nixs global excludes so the
  formatter cannot fight the Ent generator or invalidate atlas.sum.

Marks tasks 13.1 and 13.2 complete in the migrate-to-ent-and-atlas
change.

PR #1251, opened 2026-05-24
fix: address unresolved review comments from merged PRs #1179-#1246

kalbasit:
## Summary

Consolidation branch addressing unresolved reviewer threads left open across
merged PRs #1179–#1246. Each fix has its own atomic commit referencing the
source PR and thread ID; all threads are resolved on GitHub after each change.

### Bug fixes

- **cache:** touch `nar_file` by fetched row ID, not by narURL fields —
  `getNarFileFromDB` can return a `compression=none` fallback whose key differs
  from `narURL`, causing stale `last_accessed_at` and LRU regressions (#1206)
- **cache:** fail fast on `nil` eager-loaded chunk edges in `streamCompleteChunks`
  and `streamProgressiveChunks` instead of silently `continue`-ing, which caused
  count mismatches or infinite re-streaming (#1207)
- **cache:** include `narURLsToRemove` in the LRU early-return guard so orphaned
  `NarFile` blobs are passed to `parallelDeleteFromStores` instead of being
  silently leaked in storage (#1215)
- **cache:** log a warning when LRU candidate fetch hits the `maxFetchRows` cap
  so operators know eviction scanned only a partial window (#1206)
- **cache/relink:** lower `relinkBatchSize` from 500 → 499 to stay under
  SQLite's 999-placeholder limit (500 rows × 2 binds = 1000 > 999) (#1208)
- **database:** close `*sql.DB` in `Open()` when `NewClient` fails (#1208)
- **migrate/adopt:** use `ErrUnknownState` (not `ErrUnknownDialect`) for
  unrecognised `State` values — fixes false-positive `errors.Is` checks (#1188)
- **generate-migrations:** `parseSkip` validates dialect names, returning an error
  on unknown values like `"myslq"` instead of silently skipping them (#1187)
- **generate-migrations:** `--sql-only` path now calls `ensureAtlasSum` after
  writing the stub so `atlas.sum` stays in sync (#1187)
- **generate-migrations:** `runDialect` now returns the generated file path and
  prints it to stdout; `main()` warns when a second boundary falls between
  dialect runs so callers know to re-run for a consistent timestamp set (#1215)
- **generate-migrations:** add `"empty"` to `validateName` reject list (#1185)
- **nix/ent-lint-check:** add `set -o pipefail` so `ent-lint | tee` propagates
  a non-zero exit through the pipeline (#1221)
- **e2e:** fix docstring path (`var/ncps/e2e-results/` → `.e2e-results/`);
  `mc mb` and `redis-cli flushall` now use `check=True` so reset failures
  abort the test immediately (#1233)
- **e2e:** `stop_ncps` now verifies the port closes within timeout; `wait_ready`
  exits early if the child process dies before becoming ready (#1233)
- **run.py:** guard `dbmate` usage with a `shutil.which` check so a missing
  binary fails fast with a clear message instead of an obscure error (#1234)
- **CI:** pin all external GitHub Action refs to immutable commit SHAs to
  prevent supply-chain mutation (#1223)
- **CI:** remove `release-*` from pull_request/push triggers; add `go.mod` and
  `go.sum` to the ent paths-filter (#1235)
- **docs:** replace distroless-incompatible `wget` healthcheck in Docker Compose
  example with `garage status` (#1223)

### Test fixes

- Replace non-fatal `assert.Len` → `require.Len` before indexing `stamps[0..2]`
  in `generate-migrations` tests (#1185)
- Add 60 s timeout to the one-time `go build` in the test setup (#1241)
- Fix `cache_prefetch_test.go` to look up `nar_file` by `(hash, compression,
  query)` instead of hard-coded `id=1` (#1218)
- Apply `NarSize` fallback when `FileSize == 0` in `testhelper/sqlite.go` to
  avoid impossible `file_size=0` fixture rows for `compression=none` narinfos (#1218)
- Increase `chunkingDelay` from 500 ms → 1 s in `cdc_test.go` to reduce
  flakiness on slow CI machines (#1217)
- Replace `Lock`-with-timeout with `TryLock` in `cache_distributed_test.go` to
  avoid spurious failures when the lock-hold window is tight (#1217)

### Docs / skills

- Quote PostgreSQL DSNs in shell-command examples to prevent `?`-glob expansion
  by shells (#1234)
- `migrate-new/SKILL.md`: ent-lint checks A1, A2, A4 — not all five invariants
  as previously stated (#1225)
- `migrate-down/SKILL.md`: clarify that `DROP COLUMN` is sensitive/forbidden DDL
  per the migration spec and is only permissible as the terminal expand-contract
  step; add the missing "deploy code that always writes non-null" step to the
  NOT NULL recipe (#1225)
- `openspec/changes/archive/2026-05-23-less-tests/`: fix stale paths, add `bash`
  fence tag, replace "branch coverage" with "statement/block coverage" (#1245)
- `openspec/specs/test-suite-efficiency/spec.md`: same coverage terminology
  fix (#1245)
- `openspec/changes/archive/2026-05-21-migrate-to-ent-and-atlas/design.md`:
  add language tag to decision-tree fence block; fix grammar (#1186)
- `openspec/changes/archive/2026-05-21-migrate-to-ent-and-atlas/tasks.md`:
  update §10.1/§11.1 notes to reflect that `*database.Client` threading was
  completed in §11, not deferred (#1191, #1192)

## Test plan

- [x] `nix flake check` (includes all unit + integration tests, lint, ent-lint,
  atlas-sum, helm-unittest, schema-equivalence)
- [x] `go test -race ./...` locally with `eval "$(enable-integration-tests)"`
- [x] Verify all referenced GitHub threads are marked resolved

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.agent/skills/migrate-down/SKILL.md
+++ b/.agent/skills/migrate-down/SKILL.md
@@ -23,7 +23,9 @@ new binaries can coexist against the same schema.
 1. **Backfill** the new column in migration N+1 (use `--sql-only` —
    `go run ./cmd/generate-migrations --sql-only --name=backfill_x`).
 1. **Switch reads** to the new column in the application code; deploy.
-1. **Drop** the old column in migration N+2.
+1. **Remove** the old column in migration N+2. `DROP COLUMN` is flagged
+   as sensitive DDL by the migration spec and is only permissible at
+   this final step, once no deployed binary references the old column.
 
 Each migration ships in its own release; the application gracefully
 handles the dual-column state in between.
@@ -34,7 +36,9 @@ The specific case of adding a NOT NULL constraint to an existing
 nullable column:
 
 1. **Migration A**: ADD COLUMN nullable, default-able.
-1. **Migration B** (`--sql-only` stub): BACKFILL existing rows.
+1. **Deploy** application code that always writes a non-null value for
+   the new column (so all rows written after this deploy have a value).
+1. **Migration B** (`--sql-only` stub): BACKFILL existing null rows.
 1. **Migration C**: ADD CONSTRAINT NOT NULL (or
    `ALTER COLUMN ... SET NOT NULL`).
 1. **Deploy each step independently**; never combine into a single
--- a/.agent/skills/migrate-new/SKILL.md
+++ b/.agent/skills/migrate-new/SKILL.md
@@ -90,8 +90,9 @@ backfills and the four-step NOT NULL recipe.
 
 - `task ent:check` (or `go generate ./ent/... && git diff --exit-code ./ent/`)
   confirms the Ent client is up to date.
-- `go run ./cmd/ent-lint --root .` checks every schema against the five
-  codegen invariants.
+- `go run ./cmd/ent-lint --root .` checks schemas against the
+  implemented invariants (A1, A2, A4); A3 and A5 are enforced by code
+  review until wired into the linter.
 - `go run ./cmd/atlas-sum-check --root .` confirms every `atlas.sum`
   matches the directory contents.
 - `go test -race -run TestSchemaEquivalence ./migrations/...` runs the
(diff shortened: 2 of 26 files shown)


kalbasit:

- #1251 ◀

Issue #1252, opened 2026-05-24
Hang in TestCacheBackends/.../ConcurrentDownloadCancelOneClientOthersContinue — io.ReadAll has no timeout guard


kalbasit:
Spun off from #1247 after triage during the lean-flake-check work
(PR #1249).

## Observed failure

```
panic: test timed out after 20m0s
        running tests:
                TestCacheBackends/SQLite#01/ConcurrentDownloadCancelOneClientOthersContinue (19m39s)
```

Surfaced in the `ncps-s3-tests` cohort during a measurement run; the test is in `TestCacheBackends`'s SQLite subtest path, so it runs in every cohort (env-var-driven). 20-minute hang means it ran until the outer Go-test-package timeout fired.

## Why this is different from the other #1247 flakes

The other four flakes (`PutNarInfoDeadlock`, `nar_does_exist…last_accessed_at`, `GetNarInfoConcurrentPutNarInfoDuringMigration`, `TestCDCBackends/.../first_pull_completes_before_CDC_chunking_finishes`) are all wall-clock tolerance assertions racing CI load. They fail fast with clear assertion errors that say "elapsed was X, expected Y." Tractable test-only fixes.

This one is a true hang. Looking at the test body in `pkg/cache/cache_test.go:1747`, the test has `time.After` guards for every coordination point up to line 1905:

| Line | Guard |
| ---:| --- |
| 1813 | `time.After(5 * time.Second)` waiting for upstream health check |
| 1870 | `time.After(5 * time.Second)` waiting for `slowNarServed` |
| 1886 | `time.After(5 * time.Second)` waiting for client A to return after cancellation |
| 1894 | `time.After(10 * time.Second)` waiting for `downloadComplete` |
| 1903 | `time.After(5 * time.Second)` waiting for client B's `doneB` |

All of these `t.Fatal` on timeout. Since the test hit the **outer** 20m timeout instead of any of these `t.Fatal`s, the hang must be **after line 1905** — which leaves two un-guarded blocking calls:

1. **Line 1910**: `bodyB, err := io.ReadAll(readerB)` — no `context` or timeout. If the `readerB` returned by `c.GetNar` never EOFs (because the underlying download stalled or the streaming goroutine deadlocked), this blocks forever.

[...]

PR #1249, opened 2026-05-24
refactor(nix): per-backend test cohorts + lean ncps build

kalbasit:
## What this PR actually delivers

A structural rework of `nix flake check`. **CI build-job wall-clock is roughly unchanged from main** (~15m → ~15m); the wins are everywhere except total CI time:

| Thing | Before | After |
| --- | ---:| ---:|
| `nix build .#ncps` (lean binary build, no tests) | ~12m | **~1m** |
| `Build coverage` step in CI (`nix build .#ncps.coverage`) | ~6–12m | **~3s** |
| Flake check step in CI (was monolith, now cohorts) | ~8m | ~14m |
| Backend test failures localized to named cohorts | no | yes |
| Coverage piggybacked on per-cohort `cover.out` | no | yes |

The flake-check step itself got **slower**, not faster — splitting the monolith into 5 cohort derivations replaces 1 compile + 1 test run with 5 compile + 5 test runs, and the runner doesn't have enough cores to parallelize that away. The coverage piggyback (cohorts emit `cover.out`, ncps-coverage merges them) eats most of that slowdown back. Net CI: roughly the same total.

I attempted to claim "~12m → ~5m" earlier; that was wrong and I owe an apology. The 5m figure was a local-machine measurement of `nix flake check` in isolation, not CI's build job which also runs the coverage and OCI steps sequentially after. The honest CI-total number stayed flat.

## What's genuinely better

- **`nix build .#ncps` is now ~1 min** instead of ~12. Useful for every downstream that does `nix run github:kalbasit/ncps` or builds Docker images.
- **`ncps-coverage` is a 30s Python merger** of per-cohort `cover.out` instead of a 12m monolith re-run. Codecov gets the same merged profile via the same `nix build .#ncps.coverage` invocation; the symlink name (`result-coverage`) and payload shape are preserved verbatim, no upstream workflow changes needed.
- **Per-backend failure isolation.** When MariaDB flakes, only `ncps-mysql-tests` fails. CI logs name it directly instead of burying it in a 12m monolith log.
- **`packages._ncps-base`** factored out so cohorts can depend on a shared buildGoModule without cycling through `packages.ncps.passthru.coverage`.
- **`checks` attrset is explicit** — no longer pulls in `docker`, `docker-dev`, `push-docker-image`, `k8s-tests`, `update-cu-base`, `deps` as side-effect "checks" that aren't quality gates.

## Side effect: surfaced existing `pkg/cache` flakes

[...]

Issue #1247, opened 2026-05-23
Flaky test: TestCacheBackends/MySQL#01/PutNarInfoDeadlock — context deadline exceeded


kalbasit:
Observed on `main` at 61405fb5 while running `nix build .#checks.x86_64-linux.ncps -L --no-link` locally as part of the `lean-flake-check` baseline measurement.

## Failure

```
--- FAIL: TestCacheBackends (0.00s)
    --- FAIL: TestCacheBackends/MySQL#01 (190.31s)
        --- FAIL: TestCacheBackends/MySQL#01/PutNarInfoDeadlock (23.91s)
            cache_test.go:847:
                  Error Trace:    /build/source/pkg/cache/cache_test.go:847
                  Error:          Received unexpected error:
                                  error storing in database: error upserting nar_file record: context deadline exceeded
                  Test:           TestCacheBackends/MySQL#01/PutNarInfoDeadlock
                  Messages:       PutNarInfo should not deadlock or timeout
FAIL
coverage: 72.9% of statements
FAIL    github.com/kalbasit/ncps/pkg/cache      277.161s
```

The MariaDB subtest hit the upsert deadline (23.91s) on `nar_file`. The test is explicitly labelled "should not deadlock or timeout", so the failure mode is the very contract the test asserts.

## Environment

- Commit: 61405fb5d3c9cf1bd3130d32f9ad1fc7853a4218 (main)
- Backend: MariaDB (in-sandbox via `nix/packages/ncps/pre-check-mysql.sh`)
- Other backends running concurrently in the same derivation: Garage, PostgreSQL, Redis

## Reproduction

```
nix build .#checks.x86_64-linux.ncps -L --no-link
```

## Notes

- This is a pre-existing flake on `main`, not introduced by any in-flight branch.
- The race detector is enabled (`-race`) but the trace points to a context deadline rather than a data race.
- Worth investigating whether the upsert is genuinely deadlocking under contention from the other concurrent backends in the same `preCheck`, or whether the deadline (currently 23.91s expiry) is just too tight for MariaDB under load.
- Recording here so the `lean-flake-check` topology work can proceed without conflating its measurements with this flake.

kalbasit:
Second flake observed on the same MariaDB cohort during the lean-flake-check Phase 4 measurement (same `pkg/cache` package as the original, different test, different failure mode):

```
[...]

PR #1253, opened 2026-05-24
test(cache,cdc): stop racing wall-clock under CI load

kalbasit:
## Summary

Three pre-existing `pkg/cache` flakes from #1247 share one root cause: assertions that race against absolute wall-clock time and break under CI load even when the production code is correct. Surfaced repeatedly during the lean-flake-check measurement runs in #1249 (which sits stacked on top of this branch).

All three fixes are **test-only**. No production code is touched.

## What changed

| Flake (from #1247) | Test | Before | After |
| --- | --- | --- | --- |
| #5 | `TestCDCBackends/{SQLite,MySQL}/first_pull_completes_before_CDC_chunking_finishes` | `assert.Less(elapsed, 500ms)` racing the slow chunker | Happens-before check: `select { case <-chunkingStarted: t.Fatal(…) }` after `io.ReadAll`. `slowChunker` grew an optional `started chan struct{}` closed when its delay expires. Delay bumped 500 ms → 10 s (only matters under regression). |
| #1 | `TestCacheBackends/MySQL/PutNarInfoDeadlock` | `context.WithTimeout(2s)` then `assert.NoError`. Could not distinguish "deadlock" from "MariaDB slow." | Timeout bumped 2 s → 60 s. A real deadlock still fails fast; non-deadlocked slow path now passes. |
| #2 | Six instances of `assert.WithinDuration(CreatedAt, LastAccessedAt, 2s)` across `TestCacheBackends` | 2 s tolerance was tight enough that two CLOSE-TO-NOW() timestamps from different SQL operations under MariaDB load could fall outside it (observed -3 s spread). | Tolerance bumped 2 s → 30 s. Still tight enough to catch a genuinely broken "set both timestamps at row creation" semantic. |

## Why "test-only" and not "fix the production bug"

These three are *measurement bugs*: the assertions encode an assumption ("operation X completes within Y milliseconds") that's an artifact of the dev host's speed, not a property the codebase ever promised. The behavior the tests *want* to verify (no deadlock, timestamps set close together at creation, GetNar doesn't block on chunking) is still being verified — the encoding is just no longer wall-clock-dependent.

Flakes #3 and #4 from #1247 are NOT addressed here because they are not the same shape:

- **#3** (`ConcurrentDownloadCancelOneClientOthersContinue`, 20-min hang) is a real hang in `io.ReadAll(readerB)` or `wg.Wait()` after all the existing 5–10 s guards have passed — almost certainly a production-code bug in the shared-reader-fanout path. Filed as **#1252** for separate investigation.
- **#4** (`GetNarInfoConcurrentPutNarInfoDuringMigration`) exercises the same concurrent-migration-vs-PutNarInfo window that the next test (`testGetNarInfoMultipleConcurrentPutsDuringMigration`) explicitly `t.Skip`s with `"blocked on pgx transaction-abort cascade in PutNarInfo re-fetch path"`. That tracked production bug is the actual root cause; this test exercising it less reliably isn't a separate issue.

## Stack

This PR is stacked under #1249 (`improve-flake-check-time`); landing this first should unblock #1249's CI by eliminating three of the five flakes that were causing intermittent reds.

```
main → fix-cache-test-timing-assertions (this PR, #1253)
     → improve-flake-check-time (#1249)
```

## Test plan

- [x] `nix develop -- go test -race -count=1 -run "TestCDCBackends/SQLite/first_pull_completes_before_CDC_chunking_finishes" ./pkg/cache/` passes locally.
- [x] `nix develop -- go test -race -count=1 -run "TestCacheBackends/SQLite/(PutNarInfoDeadlock|GetNarInfo/narinfo_exists_upstream/nar_does_exist)" ./pkg/cache/` passes locally.
- [x] `golangci-lint run` clean (run inside `nix develop` so the toolchain resolves).

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -581,7 +581,7 @@ func testGetNarInfo(factory cacheFactory) func(*testing.T) {
 
 				require.NoError(t, err)
 				assert.Equal(t, testdata.Nar2.NarInfoHash, nim.Hash)
-				assert.WithinDuration(t, nim.CreatedAt, nim.LastAccessedAt.Time, 2*time.Second)
+				assert.WithinDuration(t, nim.CreatedAt, nim.LastAccessedAt.Time, 30*time.Second)
 			})
 
 			t.Run("nar does exist in the database, and has initial last_accessed_at", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


kalbasit:

- #1253 ◀
    - #1249

PR #1255, opened 2026-05-24
fix(cache): fix compression mismatch in CDC lazy-chunking

kalbasit:
## Root cause

In lazy CDC deployments, narinfos are initially stored with upstream
compression (e.g. `Compression: xz`). Chunking happens in the
background; `migrateNarToChunksCleanup` updates the narinfo in the DB
to `Compression: none` asynchronously. Two races cause "input
compression not recognized" on the Nix client:

**Race 1 — GetNarInfo returns stale xz narinfo while NAR is already
chunked.** During the window between chunking completing and the DB
update, `GetNarInfo` returns a narinfo with `Compression: xz`. When Nix
downloads the NAR, `GetNar` detects it is in chunks (via the
`Compression=none` CDC fallback in `getNarFileFromDB`) and serves raw
uncompressed bytes — which Nix cannot decode as xz.

**Race 2 — Nix client has locally cached the old xz narinfo.** Even
after the DB is updated, Nix clients that cached the old `.nar.xz`
narinfo request the `.nar.xz` URL directly without re-fetching the
narinfo. `serveNarFromStorageViaPipe` was routing these requests to
chunks anyway (via the CDC fallback), again serving raw bytes to a
client expecting xz data.

## Fixes

**Fix 1 (`GetNarInfo`):** After triggering background migration for a
narinfo with non-none compression in CDC mode, call `HasNarInChunks`.
If the NAR is already chunked, normalize URL and Compression to `none`
in-memory before returning. Extracted into `maybeCDCNormalizeNarInfoURL`
to keep nesting within lint limits. Narinfo fingerprints cover
StorePath/NarHash/NarSize/References — not URL or Compression — so no
re-signing is needed.

**Fix 2 (`serveNarFromStorageViaPipe`):** Only promote `hasInStore` to
`serveFromChunks` when the request itself is for `Compression=none`.
For xz requests where only chunks remain (whole-file deleted), return
`storage.ErrNotFound` so Nix falls back to an upstream cache that still
has the original compressed file. This handles stale client-side
narinfo caches regardless of what the DB says.

## Test plan

- `TestCacheBackends/*/GetNarInfo_CDCNormalizesCompressionWhenChunked` — Fix 1: narinfo normalized when chunks exist
- `TestCacheBackends/*/GetNarInfo_CDCNoNormalizationWhenNotChunked` — Fix 1: no normalization before chunking
- `TestCacheBackends/*/GetNarInfo_CDCDisabledNoNormalization` — Fix 1: no normalization without CDC
- `TestCacheBackends/*/GetNar_CDCXzRequestReturnsErrWhenChunked` — Fix 2: ErrNotFound for xz request when only chunks exist
- `TestCacheBackends/*/GetNar_CDCXzServesFromStoreWhenBothExist` — Fix 2: xz served from whole-file during lazy-chunking transition

### DIFF
--- a/openspec/changes/archive/2026-05-24-fix-compression-not-recognized/.openspec.yaml
+++ b/openspec/changes/archive/2026-05-24-fix-compression-not-recognized/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-05-24
--- a/openspec/changes/archive/2026-05-24-fix-compression-not-recognized/design.md
+++ b/openspec/changes/archive/2026-05-24-fix-compression-not-recognized/design.md
@@ -1,0 +1,97 @@
+## Context
+
+ncps supports two NAR storage modes:
+
+1. **Whole-file**: NARs stored as compressed files (e.g., `.nar.xz`) in the storage
+   backend. Narinfos reference the original compression.
+2. **CDC (content-defined chunking)**: NARs are split into content-addressed chunks,
+   always stored **uncompressed**. Narinfos should reference `Compression: none`.
+
+When **lazy CDC chunking** is enabled, narinfos fetched from upstream are initially
+stored with the upstream's compression (e.g., `Compression: xz`). This allows the NAR
+to be served from the original compressed file while chunking happens in the background.
+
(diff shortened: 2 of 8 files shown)


kalbasit:

- #1255 ◀

PR #1257, opened 2026-05-24
test(cache): fail fast on ConcurrentDownloadCancel hang (refs #1252)

kalbasit:
## Summary

testConcurrentDownloadCancelOneClientOthersContinue had two un-timeout-guarded blocking calls after the existing 5–10 s `t.Fatal` guards: `io.ReadAll(readerB)` (line 1916) and `wg.Wait()` (line 1932). When the shared streaming goroutine got stuck in some race with client A's cancellation, `readerB`'s pipe writer never closed and `io.ReadAll` blocked for the full `go test -timeout` budget — observed as **20-minute hangs** on CI under heavy load during the lean-flake-check measurement runs.

This PR wraps both calls in a new test-only helper `runWithTimeout` (30 s budget, dumps all goroutines via `runtime.Stack` on timeout) so the next occurrence of the hang fails the test in 30 s with the exact stuck-goroutine trace instead of consuming 20 minutes of CI.

## What this PR does NOT do

It does not fix the underlying race. The race is non-trivial: not reproducible locally even under 200× `-race -count` stress, code inspection in `c.GetNar`'s shared-reader-fanout path is suggestive but ambiguous (see #1252 for the hypotheses), and a real fix needs goroutine-level diagnostics from an actual CI hang.

Once this PR is merged and the hang fires once more in CI, the failure will print exactly which goroutine is stuck where. That data should let us pinpoint and fix the root cause cheaply.

## Stack

```
main → fix-narinfo-distributed-coordination-race (#1256, the real ds.stored race fix)
     → fail-fast-concurrent-download-cancel-hang (this PR)
```

## Test plan

- [x] `go test -race -count=50 -run "TestCacheBackends/SQLite/ConcurrentDownloadCancelOneClientOthersContinue" ./pkg/cache/` → 50/50 pass (the bug doesn't reproduce locally; this is purely defensive instrumentation).
- [x] `golangci-lint run ./pkg/cache/cache_test.go` → 0 issues after `--fix`.

Refs #1252.

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -1972,12 +1972,25 @@ func runWithTimeout(t *testing.T, d time.Duration, what string, fn func()) {
 		fn()
 	}()
 
+	timer := time.NewTimer(d)
+	defer timer.Stop()
+
 	select {
 	case <-done:
-	case <-time.After(d):
-		buf := make([]byte, 1<<20)
-		n := runtime.Stack(buf, true)
-		t.Fatalf("%s did not complete within %s; goroutine dump:\n%s", what, d, buf[:n])
+	case <-timer.C:
+		buf := make([]byte, 64<<10)
+		for {
+			n := runtime.Stack(buf, true)
+			if n < len(buf) {
+				buf = buf[:n]
+
+				break
+			}
+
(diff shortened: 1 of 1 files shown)


kalbasit:

- #1256
    - #1257 ◀

PR #1258, opened 2026-05-24
fix(migrations): reliably sync identity sequences on existing DBs

kalbasit:
## Summary

- Rewrite the `20260520212038_postgres_serial_to_identity.sql`
  migration to use `ALTER TABLE ... ALTER COLUMN id RESTART WITH
  (MAX(id)+1)` inside a DO block instead of
  `setval(pg_get_serial_sequence(...))` + `DROP SEQUENCE IF EXISTS`.
- The old approach had two failure modes on existing databases: the
  sequence lookup could target the wrong sequence when PostgreSQL
  reused the old BIGSERIAL sequence object for the IDENTITY column,
  and the DROP could silently drop the identity sequence itself,
  leaving it at its initial low value.
- The `RESTART WITH` variant targets the identity sequence directly
  regardless of its internal name, and wraps the DROP in an
  `EXCEPTION WHEN dependent_objects_still_exist` handler so the
  identity is preserved if PostgreSQL reused the old sequence.
- Regenerated `migrations/postgres/atlas.sum` to match.

## Why

On databases migrated from BIGSERIAL, the identity sequence was left
at its initial value (1) rather than at `MAX(id)+1`. Subsequent
auto-inserts generated primary-key values that collided with existing
rows, producing `duplicate key value violates unique constraint
"narinfos_pkey"` errors and HTTP 500 responses to Nix clients.


kalbasit:

- #1259
    - #1260
        - #1261
            - #1262
                - #1263
                    - #1258 ◀

PR #1259, opened 2026-05-25
fix: use DO NOTHING for chunk inserts, add 25P02 recovery

kalbasit:
## Summary

- Replace per-chunk `UpsertOne(...).Update(...)` with
  `Create().OnConflictColumns(hash).Ignore().Exec()` so that a
  duplicate-hash conflict leaves the transaction alive instead of
  aborting it with SQLSTATE 25P02.
- After each INSERT, query the chunk ID by hash for the
  `nar_file_chunks` link — `DO NOTHING` returns no rows so an
  explicit SELECT is required.
- Add `database.IsAbortedTransactionError` (SQLSTATE 25P02 detector)
  and a best-effort bare `ROLLBACK` in `WithTransaction` to recover
  connections that entered the aborted state before this fix was
  deployed.
- Regression coverage via three new `testRecordChunkBatch_*` subtests.

## Why

A duplicate chunk hash caused the DB engine to abort the whole
transaction. Every subsequent statement on that connection failed
with "current transaction is aborted, commands ignored until end of
transaction block", which surfaced as HTTP 500 to Nix clients.

### DIFF
--- a/openspec/changes/archive/2026-05-24-fix-postgres-aborted-transaction-500/.openspec.yaml
+++ b/openspec/changes/archive/2026-05-24-fix-postgres-aborted-transaction-500/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-05-24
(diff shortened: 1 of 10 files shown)


kalbasit:

- #1259 ◀
    - #1260
        - #1261
            - #1262
                - #1263
                    - #1258

PR #1260, opened 2026-05-25
fix: remove non-deterministic 25P02 pool ROLLBACK

kalbasit:
## Summary

- Remove the best-effort bare `ROLLBACK` that was added in the
  previous commit to recover aborted-transaction connections.
- The root cause (duplicate chunk conflicts) is eliminated by the
  `DO NOTHING` fix, so the ROLLBACK is no longer needed.
- A meaningful integration test for this behaviour is non-deterministic
  without connection-level control over `database/sql`, so the code
  was removed rather than left in with a misleading comment.

## Why

The bare `ROLLBACK` was a defensive measure for connections already
in the aborted state. Now that no statement ever aborts the transaction,
the fallback is dead code that adds complexity without benefit.

### DIFF
--- a/pkg/database/client.go
+++ b/pkg/database/client.go
@@ -140,17 +140,6 @@ func (c *Client) WithTransaction(
 			return fmt.Errorf("transaction for %s failed (rollback also failed: %w): %w", name, rbErr, err)
 		}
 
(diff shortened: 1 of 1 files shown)


kalbasit:

- #1259
    - #1260 ◀
        - #1261
            - #1262
                - #1263
                    - #1258

PR #1261, opened 2026-05-25
refactor(cache): bulk-fetch chunks before insert

kalbasit:
## Summary

- Pre-fetch all existing chunks for the current NAR in a single
  query before the per-chunk insert loop.
- Use the fetched map to skip `INSERT ... ON CONFLICT DO NOTHING`
  for chunks that are already in the DB, avoiding unnecessary round
  trips to the database.

## Why

The per-chunk query-after-insert pattern introduced in the previous
commit works correctly but issues N×2 round trips per NAR. Bulk
pre-fetching reduces this to 1 read + (new chunks only) writes,
improving throughput for NARs with many chunks.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -147,6 +147,8 @@ var (
 
 	errMissingChunkEdge = errors.New("nar_file_chunk is missing eager-loaded chunk edge")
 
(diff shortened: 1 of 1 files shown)


kalbasit:

- #1259
    - #1260
        - #1261 ◀
            - #1262
                - #1263
                    - #1258

PR #1262, opened 2026-05-25
fix(cache): avoid 25P02 on concurrent narinfo insert

kalbasit:
## Summary

- Replace the narinfo upsert `OnConflictColumns(hash).Update(...)`
  with `OnConflictColumns(hash).Ignore().Exec()` so that a
  concurrent insert of the same hash does not abort the transaction.
- After the `DO NOTHING` insert, query the row by hash to obtain the
  ID for downstream operations.

## Why

The same duplicate-key / aborted-transaction problem that affected
chunk inserts also affects narinfo inserts when two goroutines race
to cache the same hash. Applying `DO NOTHING` here closes that
remaining 25P02 surface.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -4441,27 +4441,23 @@ func upsertNarInfoFromParsed(
 
 	switch {
 	case ent.IsNotFound(err):
-		// Insert new
+		// Insert new, ignoring a concurrent insert racing between our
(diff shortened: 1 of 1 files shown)


kalbasit:

- #1259
    - #1260
        - #1261
            - #1262 ◀
                - #1263
                    - #1258

PR #1263, opened 2026-05-25
fix(cache): fill stub narinfo from concurrent race

kalbasit:
## Summary

- After the `DO NOTHING` + re-fetch pattern, detect whether the
  fetched row is a stub (URL is NULL or empty) written by a
  concurrent goroutine mid-flight.
- If so, issue an `UpdateOneID` to fill in the caller's narinfo data
  before returning the row.

## Why

`DO NOTHING` returns the pre-existing row as-is. If another goroutine
inserted a stub row (URL=NULL) just before us and hasn't filled it
yet, callers would operate on an incomplete narinfo record. Detecting
and updating the stub ensures the row is always fully populated after
`upsertNarInfoFromParsed` returns.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -4460,6 +4460,18 @@ func upsertNarInfoFromParsed(
 			return nil, fmt.Errorf("error fetching narinfo record after insert for hash %q: %w", hash, err)
 		}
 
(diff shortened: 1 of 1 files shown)


kalbasit:

- #1259
    - #1260
        - #1261
            - #1262
                - #1263 ◀
                    - #1258

PR #1267, opened 2026-05-25
fix(helm): migration job must not mount storage or tmp for non-SQLite databases

kalbasit:
## Summary

- The migration Job was mounting the storage PVC and an 8 GiB in-memory `tmp` emptyDir whenever local storage was configured, even for PostgreSQL/MySQL deployments. `ncps migrate up` only opens a database connection — it never reads or writes to either volume.
- Remove `tmp` volume and volumeMount from the migration Job unconditionally.
- Scope the `storage` volume and volumeMount to `database.type == sqlite` only; non-SQLite databases need no filesystem path.
- Gate the `volumes:` block itself so non-SQLite Jobs render no `volumes:` key in the Pod spec.

## Test plan

- [x] New helm-unittest cases assert no volumes for PostgreSQL + local-storage and MySQL + local-storage migration Jobs
- [x] Regression guard asserts SQLite still gets the storage PVC and volumeMount
- [x] `helm unittest charts/ncps` — 138 tests pass
- [x] `nix flake check --no-build` evaluates cleanly

### DIFF
--- a/charts/ncps/templates/migration-job.yaml
+++ b/charts/ncps/templates/migration-job.yaml
@@ -98,6 +98,6 @@ spec:
       volumes:
         - name: storage
           persistentVolumeClaim:
-            claimName: {{ include "ncps.fullname" . }}
+            claimName: {{ .Values.config.storage.local.persistence.existingClaim | default (include "ncps.fullname" .) }}
       {{- end }}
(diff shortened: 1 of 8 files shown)


kalbasit:

- #1267 ◀

PR #1268, opened 2026-05-25
fix(fsck): batch IN-clause queries to avoid Postgres cap

kalbasit:
ncps fsck aborted in phase 1g on production-sized PostgreSQL caches
with `extended protocol limited to 65535 parameters`. The CDC
size-mismatch query loaded every CDC nar_file in one shot and asked
Ent to fan-out load the linked narinfo_nar_files rows, which compiled
to `WHERE nar_file_id IN ($1...$N)`. Once N exceeded 65535 the driver
refused the statement and fsck never reached phase 2. The same shape
in chunksForNarFile would have failed for any single CDC NAR with
> 65535 chunks via the `WithChunk()` eager-load.

Replace both eager-load patterns with keyset-paginated walks bounded
by a shared `fsckEagerLoadBatchSize = 1000` constant: walk the parent
table by ID/chunk_index in pages, then issue a bounded
`NarFileIDIn(...)` (size-mismatch) or `IDIn(...)` (chunk-walk) lookup
per page. Behavior is unchanged on small caches; both call sites keep
their existing signatures and ordering contracts.

Add two PostgreSQL-cohort regression tests gated on
NCPS_TEST_ADMIN_POSTGRES_URL that seed > 70_000 rows via CreateBulk
and confirm the function paths succeed. Verified that both tests
reproduce the production error against the pre-fix code.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/openspec/changes/archive/2026-05-25-fsck-is-failing/.openspec.yaml
+++ b/openspec/changes/archive/2026-05-25-fsck-is-failing/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-05-25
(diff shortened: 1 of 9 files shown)


kalbasit:

- #1268 ◀

PR #1273, opened 2026-05-25
test(migrate): use timestamp cutoff in filterDbmateEra

kalbasit:
The previous implementation filtered bridge-era migrations by name
pattern ("ent_baseline", "serial_to_identity"). Adding any new
post-bridge migration caused the dbmate-era seed test to break with
a "missing out-of-order migration" error from goose.

Switch to a timestamp-based cutoff (bridgeEraStart = 20260520000000).
Any migration with a version >= that constant is post-bridge and is
excluded from the dbmate-era seed, regardless of its name. This
matches the existing fileVersions() parsing approach and will
correctly exclude all future incremental migrations without requiring
a name-list update.


kalbasit:

- #1275
    - #1273 ◀
        - #1272

PR #1279, opened 2026-05-28
fix(cache): prevent spurious narinfo purge under concurrent fetch

### LINKED ISSUE #1276: Flaky test: TestGetNar_NixServeUpstream/nar_served_with_zstd_transparent_encoding — narinfo returns 404 under parallel subtest load
## Observed failure

In CI (`ncps-postgres-tests`, and likely other cohort derivations), the subtest
`TestGetNar_NixServeUpstream/nar_served_with_zstd_transparent_encoding` in
`pkg/server/server_test.go` fails intermittently with:

```
--- FAIL: TestGetNar_NixServeUpstream (0.14s)
    --- FAIL: TestGetNar_NixServeUpstream/nar_served_with_zstd_transparent_encoding (0.06s)
        server_test.go:1527:
                   Error Trace: github.com/kalbasit/ncps/pkg/server/server_test.go:1527
                   Error:       Not equal:
                                expected: 200
                                  actual: 404
                   Test:        TestGetNar_NixServeUpstream/nar_served_with_zstd_transparent_encoding
```

The failure is at the **narinfo** fetch step (line 1527), before the NAR is even
requested.  The test is fetching
`GET /{Nar7.NarInfoHash}.narinfo` and receiving 404 instead of 200.

## Context

`TestGetNar_NixServeUpstream` tests proxying from a nix-serve style upstream
(`Compression: none`, no `Content-Encoding` support).  It has three parallel
subtests that all share the **same ncps server instance and cache**:

1. `narinfo has Compression:none and FileSize==NarSize`
2. `nar served as raw bytes without Accept-Encoding`
3. `nar served with zstd transparent encoding` ← fails

All three subtests start by fetching the same narinfo (`Nar7`, `Compression:
none`).  When they run truly in parallel (as they do in the Nix sandbox build),
all three hit the narinfo endpoint simultaneously.  The suspected race: ncps
begins fetching + storing the narinfo on the first request, and concurrent
requests racing through the same code path can intermittently produce a 404
while the first fetch/store is in flight.

## Relationship to previous issues

Issues #227, #615, and #806 (all closed under milestone *v0.8.5 – Support for
nix-serve as upstream*) addressed correctness bugs with `Compression: none`
narinfos.  This is a different, narrower problem: a **concurrency / timing
issue** that only surfaces under parallel test execution in the sandbox.

The test itself was added in #1047 ("test: add transparent zstd serving
coverage") and passes reliably on `main` when the build is cached; it fails
when the derivation runs fresh (cache miss) and all subtests execute
simultaneously without prior warm-up.

## Reproduction

The failure appears in the `ncps-postgres-tests` (and probably `ncps-s3-tests`
and `ncps-mysql-tests`) Nix CI derivations.  It does not reproduce reliably with
`task test` locally because local tests share a process and the SQLite path is
more forgiving of concurrent access.

## Expected behaviour

All three subtests should independently be able to fetch the narinfo and receive
HTTP 200, regardless of parallel execution order.

## Suggested investigation

- Check whether ncps's narinfo fetch-and-store path is protected against
  concurrent requests for the same hash (i.e. does it hold a per-hash lock
  before fetching from upstream, or can two goroutines both receive "not found"
  and both attempt the upstream fetch simultaneously?).
- The prior fix in #882 addressed base16 vs nix32 hash handling; a similar
  in-flight deduplication guard may be needed for the concurrent-fetch case.

kalbasit:
## Summary

- **Root cause**: `hasUpstreamJob` only checked `narJobKey` (`"download:nar:…"`), not `narInfoJobKey` (`"download:narinfo:…"`). When concurrent goroutines piggybacked on a cold-cache narinfo fetch, they woke on `ds.done` after the narinfo was stored in the DB but before the NAR download goroutine had registered its own job key. In that narrow TOCTOU window, both `hasUpstreamJob` and `isRemoteDownloadInProgress` returned false — causing the purge guard in `getNarInfoFromDatabase` to fire, deleting the narinfo and returning `storage.ErrNotFound` → HTTP 404 to the waiting callers.
- **Fix**: Extend `hasUpstreamJob` to check `narInfoJobKey` in addition to `narJobKey`. If the narinfo job is still registered in `upstreamJobs`, the purge guard is suppressed.
- **Test**: Add `testGetNarInfoConcurrentColdCacheFetch` — 10 goroutines call `GetNarInfo` for the same hash simultaneously against a cold cache; all must succeed with no error.

Closes #1276

## Test plan

- [x] `task fmt` exits 0 with no changes
- [x] `task lint` exits 0 (`GOWORK=off golangci-lint run`)
- [x] `GOWORK=off go test -race -count=10 ./pkg/cache/...` — all pass
- [x] `GOWORK=off go test -race ./...` — all packages pass

### DIFF
--- a/openspec/changes/archive/2026-05-28-fix-flaky-narinfo-concurrent-fetch/.openspec.yaml
+++ b/openspec/changes/archive/2026-05-28-fix-flaky-narinfo-concurrent-fetch/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-05-26
--- a/openspec/changes/archive/2026-05-28-fix-flaky-narinfo-concurrent-fetch/design.md
+++ b/openspec/changes/archive/2026-05-28-fix-flaky-narinfo-concurrent-fetch/design.md
@@ -1,0 +1,71 @@
+## Context
+
+`pkg/cache/cache.go` has two layers of concurrency control for upstream fetches:
+
+1. **`coordinateDownload`** (line ~5318): takes a `narInfoJobKey` or `narJobKey`, checks `upstreamJobs`, and if no job is in-flight, acquires a distributed lock (`downloadLocker`), re-checks `hasAsset`, then starts a goroutine and registers in `upstreamJobs`. Concurrent callers for the same key piggyback onto the running job via `ds.start`/`ds.stored`/`ds.done` channels.
+
+2. **`getNarInfoFromDatabase`** (line ~4078): called by `GetNarInfo` after the narinfo is stored, to validate it. Contains a purge guard: if `HasNarInStore(narURL)` returns false AND `hasUpstreamJob(narURL.Hash)` returns false AND `isRemoteDownloadInProgress` returns false, it purges the narinfo from DB and returns `errNarInfoPurged`.
+
+The race is in `getNarInfoFromDatabase`'s purge guard. The sequence of checks is non-atomic:
+
+```
+// goroutine A (concurrent request that piggybacked)
+1. HasNarInStore(narURL) → false        // NAR rename not yet visible
+   [goroutine B's NAR download completes: os.Rename(tmp→narPath)]
+   [goroutine B's pullNarInfo deferred done() removes job from upstreamJobs, closes ds.done]
+2. hasUpstreamJob(narURL.Hash) → false  // job just removed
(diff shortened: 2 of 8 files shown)


kalbasit:

- #1279 ◀

PR #1280, opened 2026-05-28
fix(cache): prevent hang when client cancels concurrent NAR download

### LINKED ISSUE #1252: Hang in TestCacheBackends/.../ConcurrentDownloadCancelOneClientOthersContinue — io.ReadAll has no timeout guard

kalbasit:
## Summary

- The per-client streaming goroutine in `GetNar` had two blocking points with
  no context cancellation escape, causing a once-in-CI 20-minute hang when one
  client cancelled a shared in-flight download (ncps #1252).

- **`cond.Wait()` loop**: added a watcher goroutine per streaming SafeGo that
  calls `ds.cond.Broadcast()` when the client's `ctx` is cancelled, waking the
  loop immediately. Added `ctx.Err() != nil` alongside the existing
  `downloadError` check so the goroutine exits without waiting for the next
  data broadcast.

- **Storage-wait `select`**: extended both the regular and decompression path
  selects (`ds.stored` / `ds.done`) with `case <-ctx.Done():` so a cancelled
  client does not block indefinitely if storage stalls under I/O pressure.

- **`fileAvailableReader`**: added `ctx context.Context` field; `Read()` now
  returns `ctx.Err()` after any `cond.Wait()` wakeup when the context is done,
  covering the decompression streaming path.

- Added `TestConcurrentDownloadCancelStress` (50 iterations, skipped under
  `-short`) to surface the race reliably. 50/50 pass with fix under `-race`.

## Test plan

- [x] `task fmt` — no changes
- [x] `task lint` — 0 issues
- [x] `task test` — all pass including `ConcurrentDownloadCancelOneClientOthersContinue`
- [x] `TestConcurrentDownloadCancelStress` — 50/50 iterations pass under `-race`

Closes #1252

### DIFF
--- a/openspec/changes/archive/2026-05-28-fix-concurrent-download-cancel-hang/.openspec.yaml
+++ b/openspec/changes/archive/2026-05-28-fix-concurrent-download-cancel-hang/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-05-28
--- a/openspec/changes/archive/2026-05-28-fix-concurrent-download-cancel-hang/design.md
+++ b/openspec/changes/archive/2026-05-28-fix-concurrent-download-cancel-hang/design.md
@@ -1,0 +1,135 @@
+## Context
+
+`GetNar` deduplicates in-flight downloads via a shared `downloadState` (`ds`). Each caller
+gets its own `io.Pipe` and a per-client SafeGo streaming goroutine that tails the temp file and
+forwards bytes. The goroutine has two blocking points:
+
+1. **`cond.Wait()` loop** (`cache.go:~1240`): sleeps until `ds.bytesWritten` advances or
+   `ds.finalSize` is set. Exits on `downloadError` but **not** on the client's own context
+   cancellation.
+2. **Storage-wait `select`** (`cache.go:~1301`): `select { case <-ds.stored: case <-ds.done: }`.
+   Neither channel is closed until `pullNarIntoStore` completes storage and returns. If storage
+   stalls (e.g., I/O pressure on a busy CI runner), neither channel fires and the goroutine
+   blocks indefinitely — leaving `defer writer.Close()` un-run, so `io.ReadAll(readerB)` hangs.
+
+PR #1257 added a 30-second `runWithTimeout` guard with goroutine dump around the two
+previously un-guarded blocking calls in the test. The test now fails fast with diagnostics
+instead of consuming the full 20-minute `go test` timeout.
+
+## Goals / Non-Goals
+
+**Goals:**
+- Ensure the streaming goroutine for any client always terminates within a bounded time,
+  regardless of whether a sibling client cancels or storage stalls.
+- Make context cancellation an explicit exit condition in the `cond.Wait()` loop so cancelled
+  clients clean up promptly without depending on downstream broadcasts.
+- Preserve the correctness invariant: `HasNar` returns true before the streaming goroutine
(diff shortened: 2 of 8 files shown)


kalbasit:

- #1280 ◀

PR #1285, opened 2026-05-29
refactor(database): consolidate dialect mappers and ErrUnknownDialect

kalbasit:
## Summary

Post-Ent-migration cleanup. The `database.Type` → ent dialect-string mapping was duplicated: `database.EntDialectFor` (documented as shared with the migrate package) had a byte-for-byte copy `entDialectFor` in `pkg/database/migrate/fresh.go`, plus a parallel `ErrUnknownDialect` sentinel. The two goose mappers (`gooseStoreDialectFor`, `gooseDialectFor`) duplicated the same switch.

- Route the migrate package through the shared `database.EntDialectFor`; delete the duplicate (and the now-unused `entgo.io/ent/dialect` import).
- Collapse the duplicated `ErrUnknownDialect` to the single `database.ErrUnknownDialect` across `state.go`, `adopt.go`, `apply.go`.
- Merge the goose mappers: `goose.Dialect` is a type alias of goose-store's `database.Dialect`, so `fresh.go` uses the shared `gooseDialectFor`.
- Behavior-preserving; records a `database-orm` spec requirement for the single-source-of-truth invariant.

## Test plan

- [x] `task fmt` / `task lint` (0 issues) / `task test` (full suite green)
- [x] Existing `pkg/database` and `pkg/database/migrate` suites cover dialect mapping + fresh install

PR #1286, opened 2026-05-29
refactor(cache): extract repeated Ent query helpers

kalbasit:
## Summary

`pkg/cache/cache.go` (~7,500 lines) inlined the same Ent queries repeatedly: the single NarInfo-by-hash lookup at several sites against both the client and `*ent.Tx`, two chunk-by-hashes batch reads, and the total-NAR-size aggregate (`SUM(file_size)` + `sql.NullInt64` unpack) duplicated as a multi-line block at two call sites.

- Add `pkg/cache/queries.go` with three unexported helpers — `narInfoByHash`, `chunksByHashes`, `totalNarFileSize` — that take the Ent entity client (`*ent.NarInfoClient` etc.), which both `*ent.Client` and `*ent.Tx` expose, so one helper serves transactional and non-transactional callers.
- Replace only the byte-identical inline sites; eager-loaded / multi-predicate variants are left untouched.
- `totalNarFileSize` returns `(int64, error)` so each caller keeps its own error policy (metrics warns and continues; cleanup returns).
- Behavior-preserving: identical SQL and error semantics. Dropped the now-unused `database/sql` import from cache.go; added two justified gosec nolints on `uint64(narTotalSize)` (file_size is a uint64 column, so the SUM is non-negative).

## Test plan

- [x] `task fmt` / `task lint` (0 issues) / `task test` (full suite green; `pkg/cache` ~30s)
- [x] Existing cache suite covers get/insert-by-hash, chunk batch reads, and cleanup/metrics size paths

### DIFF
--- a/openspec/changes/archive/2026-05-29-extract-cache-query-helpers/.openspec.yaml
+++ b/openspec/changes/archive/2026-05-29-extract-cache-query-helpers/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-05-29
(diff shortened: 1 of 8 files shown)


kalbasit:

- #1285
    - #1286 ◀
        - #1287
            - #1288

PR #1287, opened 2026-05-29
refactor(cache): standardize not-found checks

kalbasit:
## Summary

The cache layer detected missing rows inconsistently: `pkg/cache` called `ent.IsNotFound` directly at ~20 sites while the purpose-built `database.IsNotFoundError` helper was used only once. That helper is the intended single predicate — it matches both Ent's `*NotFoundError` and the `database.ErrNotFound` sentinel returned by the cache test fakes. Calling `ent.IsNotFound` directly meant those sites silently failed to recognize the sentinel.

- Replace every direct `ent.IsNotFound(x)` in `cache.go` and `build_trace.go` with `database.IsNotFoundError(x)`, preserving each argument and negation; add the `pkg/database` import to `build_trace.go`.
- Behavior-preserving for real Ent errors (the helper is a superset matcher) and strictly more correct for the sentinel, which the cache paths now honor.
- Package-specific sentinels (`storage`/`upstream`/`chunk`/`config` `ErrNotFound`) are left matched by their own `errors.Is` checks.
- Records a `database-orm` spec requirement for the single-predicate contract.

## Test plan

- [x] `task fmt` / `task lint` (0 issues) / `task test` (full suite green)
- [x] Existing cache suite (incl. fakes returning `database.ErrNotFound`) covers the not-found paths

### DIFF
--- a/openspec/changes/archive/2026-05-29-standardize-notfound-checks/.openspec.yaml
+++ b/openspec/changes/archive/2026-05-29-standardize-notfound-checks/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-05-29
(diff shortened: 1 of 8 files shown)


kalbasit:

- #1285
    - #1286
        - #1287 ◀
            - #1288

Issue #1289, opened 2026-05-29
CDC: serve whole NAR from shared storage during chunking (+ GC after grace) instead of progressive chunk stream


kalbasit:
## Summary

When CDC is enabled, a NAR that is **actively being chunked** is served to clients via the **progressive chunk-stream** path (`streamProgressiveChunks`, `pkg/cache/cache.go:6908`) rather than from a whole NAR. In a multi-replica deployment this is the *only* cross-pod option during the chunking window, and it is both a **correctness risk** and a likely **performance regression**. This issue proposes serving a whole NAR from shared storage during the chunking window, with a background GC job that reclaims the whole copy after chunking completes plus a configurable grace period.

This is related to but distinct from #1230 (which covers the *crash-mid-chunking* orphan-state variant). This issue is about the **steady-state serve-during-chunking design**, including the multi-pod case.

## Current behavior (with references)

During chunking, the `nar_file` row has `total_chunks = 0` and a fresh `chunking_started_at`. The read path treats this as "actively chunking, stream progressively":

- `GetNar` marks the NAR as available while chunking is fresh (`pkg/cache/cache.go:1049-1059`, gated on `time.Since(*nr.ChunkingStartedAt) < cdcChunkingLockTTL`).
- `getNarFromChunks` (`pkg/cache/cache.go:6697`) branches to `streamProgressiveChunks` when `total_chunks == 0` (`~6766`).
- `streamProgressiveChunks` (`pkg/cache/cache.go:6908`) polls the DB every **200ms** (`pollInterval`), batch-queries up to 256 `nar_file_chunk` rows ordered by `chunk_index`, fetches each chunk blob from the chunk store, and waits up to **`maxWaitPerChunk = 30s`** for the next index to appear.

The whole NAR that *does* exist during download is a **node-local temp file** (`createTempNarFile` → `c.tempDir`), surfaced to same-pod readers via `fileAvailableReader` (`pkg/cache/cache.go:487`). It is deleted once chunking completes (`~1611`, `~2699-2703`). It is **never written to shared storage**, so other replicas cannot read it.

## Problems

1. **Cross-pod serving has no whole-NAR fast path.** Same-pod concurrent readers stream from the local temp file (fast). A *different* replica cannot see that temp file, so it falls back to `streamProgressiveChunks` (or, if it lost the download lock and the asset never lands, to a poll-timeout error — see the companion 500 fix). The common multi-pod case is served by the slowest, most fragile path.

[...]

PR #1290, opened 2026-05-29
fix(cache): stop returning HTTP 500 on download-lock contention

kalbasit:
## Summary

Fixes the HTTP 500s seen during concurrent `nix` pulls against a
multi-replica ncps (the `failed to acquire download lock and timeout
polling for completion` errors).

When a replica failed to acquire the distributed download lock, it fell
into a poll-only loop that watched solely for the asset to appear. If the
lock holder's download failed (e.g. an upstream HTTP/2 reset) or simply
ran past the fixed poll timeout, the waiter timed out and returned the
coordination error → HTTP 500. Nix treats 500 as a hard error and retries
it up to 5×, amplifying load during exactly the concurrent-pull bursts
that request coalescing exists to smooth.

### What changed

- Rework the cross-pod fallback in `coordinateDownload` into a
  **poll-or-reacquire** loop (extracted to `pollForDownloadOrTakeOver`):
  each tick checks for the asset **and** re-attempts lock acquisition.
  - Asset appears → holder succeeded, serve it.
  - Lock re-acquired → the previous holder finished or failed without
    producing the asset, so we **take over** as the sole downloader. The
    `upstreamJobs` map and the lock keep downloads serialized, so the
    concurrent-CDC path is never exercised.
  - Wait up to `max(downloadLockTTL, downloadPollTimeout)` so a
    legitimately slow holder is not abandoned, then give up with a clean
    cache miss (`storage.ErrNotFound` → **HTTP 404**) so Nix falls back to
    another substituter instead of retrying a 500. Caller cancellation
    still surfaces as a context error.
- Add the `ncps_download_coordination_fallback_total` counter
  (`outcome=served_by_peer|take_over|give_up|caller_canceled`) plus
  structured logs so operators can see how contention resolves.
- Also includes earlier fixes on this branch balancing `ds.wg` on
  `GetNar` early returns.

Follow-up filed: #1289 (serve a whole NAR from shared storage during CDC
chunking instead of the progressive chunk stream).

## Test plan

- [x] New deterministic mock-locker tests (RED→GREEN): take-over serves
      the NAR, narinfo miss → 404, NAR give-up → 404
      (`pkg/cache/coordination_internal_test.go`).
- [x] `task fmt` (clean), `task lint` (0 issues), `task test` pass.
- [x] `task test:auto` — full suite incl. the Redis-backed distributed
      cohort — passes.

### DIFF
--- a/openspec/changes/archive/2026-05-29-fix-download-coordination-500/.openspec.yaml
+++ b/openspec/changes/archive/2026-05-29-fix-download-coordination-500/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-05-29
--- a/openspec/changes/archive/2026-05-29-fix-download-coordination-500/design.md
+++ b/openspec/changes/archive/2026-05-29-fix-download-coordination-500/design.md
@@ -1,0 +1,173 @@
+## Context
+
+In a multi-replica deployment with a Redis lock backend, ncps coalesces duplicate
+downloads of the same NAR/narinfo hash through `coordinateDownload`
+(`pkg/cache/cache.go:5386`). Coordination has two layers:
+
+- **Intra-pod**: an in-process `upstreamJobs` map keyed by lock key. A second
+  local goroutine waits on the shared `downloadState` channels (`ds.start`,
+  `ds.stored`, `ds.done`) and can observe the holder's error via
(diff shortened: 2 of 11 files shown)


kalbasit:

- #1290 ◀

PR #1291, opened 2026-05-29
ci: run integration suite on x86_64 only, not aarch64

kalbasit:
## Why CI got slow and flaky

CI averages ~20 min, often exceeds 30, and fails roughly half of runs — so a PR takes several attempts to merge. Root cause (measured on run `26653285682`):

| `build` matrix leg | Duration | Result |
|---|---|---|
| `x86_64-linux` | 17.4 min | ✅ |
| `aarch64-linux` | 25.7 min | ❌ (Postgres cohort timeout) |

The shared `build` job fans out over `["x86_64-linux","aarch64-linux"]` and runs the **entire** `nix flake check` — all integration cohorts (Garage/Postgres/MySQL/Redis) — on **both** legs. The aarch64 leg runs on slower ARM runners where the Postgres cohort times out, making it both the long pole *and* the dominant failure. Integration tests are architecture-independent, so the aarch64 run buys no signal — only wall clock and flaky reruns.

## What this does

Passes `test_systems: '["x86_64-linux"]'` to the shared `kalbasit/gh-actions` workflow so `nix flake check` + coverage run on **x86_64 only**. The `aarch64-linux` leg still builds and pushes its OCI image (the lean, test-free `packages.ncps`), so the **multi-arch manifest is unchanged** — it just no longer re-runs the integration suite.

## ⚠️ Merge order

Depends on **kalbasit/gh-actions#8** (adds the `test_systems` input). Since this repo consumes `gh-actions/.github/workflows/ci.yml@main`, **#8 must merge first** — until then this PR's `shared` job will fail with an unknown-input error. Re-run CI here after #8 lands.

## Scope / non-goals

Per the OpenSpec change `speed-up-ci`: this kills the flakiness and removes the ~26-min aarch64 critical path. It does **not** rework the dd9622c per-backend cohort split (the secondary ~8m→~14m flake-check regression) — that's a deferred, separate lever toward the eventual ~5-min target.

OpenSpec: `openspec/changes/speed-up-ci/`.


kalbasit:

- #1291 ◀

PR #1292, opened 2026-05-29
perf(nix): compile cohort test binaries once, share via GOCACHE

kalbasit:
## Why

After #1291, the `x86_64-linux` build leg (~17m) is CI's bottleneck, and the **"Flake check" step is 16.3m of it** (measured, run 26661047200). dd9622c split flake-check into 5 cohort derivations, each of which recompiles the race+coverage-instrumented test binaries from scratch — Go's build cache isn't shared across Nix sandboxes. The **4 backend cohorts compile byte-identical binaries** (cgo `mattn/go-sqlite3` under `-race`, ~2 min apiece on CI's 4-core runners), so 3 of those compiles are pure waste that core contention can't parallelize away (this is dd9622c's 8m→14m regression).

## What

Add a `_ncps-test-cache` derivation that compiles **both** cohort test sets exactly once (`go test -run '^$'`, runs zero tests so no backend is needed) and exports the populated `GOCACHE`. Each cohort seeds a writable copy of that cache before its `go test`, turning compilation into a cache hit. Cohort isolation, per-cohort `cover.out`, the merged Codecov profile, the race detector, and `t.Skip` gating are all unchanged.

Two Nix-specific details (documented in the code):
- `disallowedReferences = [ ]` — the cache legitimately references the Go toolchain.
- `dontFixup = true` — patchelf/strip would rewrite cached objects and invalidate Go's content hashes.
- Cohorts copy the cache preserving mode then `chmod -R u+w` — `--no-preserve=mode` would strip the `covdata` helper's +x bit and break coverage.

## Verification

- `_ncps-test-cache` builds (821M cache).
- **cmd cohort passes reusing the cache** — confirmed compile reuse (a pre-fix run failed only on fork/exec of the *cached* `covdata` binary, proving compilation was served from cache, not rebuilt).
- The postgres cohort's local `TestSchemaEquivalence` failure is environmental: it fails identically with this change stashed, and the cohort passes in CI on this branch.

This PR's own CI run will give the real before/after x86-leg number.

OpenSpec: `flake-check-topology` spec gains "Race+coverage test binaries are compiled once and shared".

### DIFF
--- a/nix/checks/flake-module.nix
+++ b/nix/checks/flake-module.nix
@@ -40,6 +40,66 @@
         redis # Redis distributed-lock integration tests
       ];
 
+      # Compile-affecting `go test` config, shared VERBATIM between the
+      # _ncps-test-cache derivation (which compiles the instrumented test
+      # binaries once) and every cohort's own `go test` (which reuses them via
+      # GOCACHE). Only `-race`, `-covermode`, `-coverpkg`, the package set, and
+      # the source affect Go's build cache key; the run-only flags each cohort
+      # adds (`-count=1`, `-timeout`, `-coverprofile`) do not, so they may differ
+      # without causing a cache miss. If THESE drift between cache and cohort,
+      # the cohort silently recompiles (correct, just slow) — keep them in this
+      # single binding so the two sites can never disagree.
+      testSpecs = {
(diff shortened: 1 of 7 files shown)


kalbasit:

- #1292 ◀

PR #1293, opened 2026-05-29
ci: fan flake checks out into a parallel matrix

kalbasit:
## Why

After compile-once (#1292) CI is ~10m, bottlenecked by a single `nix flake check` building **~13 check derivations on one contended 4-core runner**. They're logically parallel but CPU-starved; the remaining costs (backend-cohort test execution, golangci's own analysis) aren't cacheable — they need more cores in parallel.

## What

Replace the monolithic check with a **CI matrix**:
- `check-matrix` job enumerates checks: `nix eval .#checks.x86_64-linux --apply builtins.attrNames --json`.
- `checks` job runs each as its own parallel runner: `nix build .#checks.x86_64-linux.<name>` (`fail-fast: false`), pulling the shared `_ncps-test-cache`/`_ncps-base` from Cachix instead of rebuilding.
- `coverage` job (`needs: checks`) merges per-cohort profiles → Codecov.
- The shared build job sets **`run_flake_check: false`** (kalbasit/gh-actions#9) so it only builds/pushes the OCI image.

Wall-clock collapses from sum-under-contention (~10m) to **slowest-single-check + nix setup (~3–4m)** — the path to the ~5m target. The per-job times on this run also give us the first clean per-check breakdown.

## ⚠️ Merge order
Depends on **kalbasit/gh-actions#9** (adds `run_flake_check`). The caller is **temporarily pinned to the gh-actions branch** to validate now; revert to `@main` before merging once #9 lands.

OpenSpec: `flake-check-topology` gains "CI runs each check as an independent parallel job" + "Coverage is produced by a job consuming the fanned-out cohorts".

PR #1294, opened 2026-05-29
ci(releases): scope flake-check to x86_64

kalbasit:
## Summary

The release workflow (`v*.*.*` tag builds) called `build.yml` **without** `test_systems`, so it re-ran the full `nix flake check` + coverage on **both** x86_64 and aarch64 — including the slower, flaky aarch64 Postgres path that the CI workflow (#1291) already learned to avoid.

- `releases.yml`: pass `test_systems: '["x86_64-linux"]'` to the shared build workflow. The integration suite + coverage now run on x86_64 only; the aarch64 leg still **builds and pushes** its OCI image, so the multi-arch manifest is unchanged. `run_flake_check` stays at its default (`true`) so releases still gate on the full suite.
- Mirrors the CI workflow's scoping.
- `flake-check-topology` spec: the "Integration cohorts validated on a single CI architecture" requirement now covers **both** `ci.yml` and `releases.yml`, with a new release-tag scenario.

Independent of the fan-out PR (#1293); branched off `main`. Relies on the `test_systems` input already on `kalbasit/gh-actions@main` (merged in #8).

## Test plan

- [x] `actionlint .github/workflows/releases.yml` — exit 0
- [x] `task fmt` clean; pre-commit hooks pass
- [ ] Confirm on the next release tag that the aarch64 leg builds the image but skips `nix flake check`

### DIFF
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -11,6 +11,12 @@ jobs:
     uses: kalbasit/gh-actions/.github/workflows/build.yml@main
     with:
       systems: '["x86_64-linux", "aarch64-linux"]'
+      # Validate the integration suite + coverage on x86_64-linux only (the
+      # cohorts are architecture-independent). The aarch64-linux leg still
+      # builds and pushes its OCI image for the multi-arch manifest, but no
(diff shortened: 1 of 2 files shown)


kalbasit:

- #1294 ◀

PR #1295, opened 2026-05-29
fix(cache): balance ds.wg on early GetNar returns

kalbasit:
## Summary

- In `GetNar`, `ds.wg.Add(1)` is called when `canStream=true` to track active
  readers, but two early-return paths never called the matching `ds.wg.Done()`:
  1. `case <-ctx.Done()` — context cancelled before `<-ds.start` closed
  2. `if err := ds.getError()` — error detected after start but before goroutine launch
- The streaming goroutine has `defer ds.wg.Done()`, but it is never launched in
  these paths, so the WaitGroup counter is never decremented.
- The cleanup goroutine added to `backgroundWG` calls `ds.wg.Wait()`, which
  blocked forever, causing `c.Close()` to hang indefinitely.
- Fix: add `ds.wg.Done()` before each of the two early returns.
- Regression test: `TestGetNarCloseDoesNotHangAfterContextCancelledBeforeStart`
  pre-cancels the context and verifies `c.Close()` completes within 10s.

## Test plan

- [x] `task fmt` passes
- [x] `task lint` passes
- [x] `task test` passes
- [x] Regression test confirmed RED before fix, GREEN after


kalbasit:

- #1295 ◀

kalbasit:
Empty PR — the ds.wg.Done fixes were already incorporated into main via PR #1290. Closing.

PR #1297, opened 2026-05-30
feat(cache): CDC recovery follow-ups (GC, backoff, streaming)

kalbasit:
## Summary

Three hardening follow-ups left over from the archived `fix-phantom-nar-cache-miss`
change (and its keyset-cursor commit), stacked on #1296. OpenSpec change
`cdc-recovery-followups` (proposal/design/specs/tasks, all 17 tasks complete,
`/opsx:verify` clean, archived).

- **Placeholder-row GC** (`feat(cache): GC genuinely-absent placeholder rows`):
  the CDC lazy-recovery sweep skipped backing-less `nar_file` placeholder rows but
  never cleared them, so rows for NARs genuinely gone upstream accumulated and were
  re-scanned every cursor cycle. New `upstream.NarInfoExistence` returns a tri-state
  `Existence{Present,Absent,Unknown}` — it probes the **narinfo** HEAD (unambiguous)
  and, unlike `HasNar`/`HasNarInfo`, does **not** collapse a timeout into "absent".
  The sweep deletes a row only when **every** healthy upstream is definitively
  `Absent` (≥1 healthy upstream, none Present/Unknown). `ON DELETE CASCADE` cleans up
  links. Present / Unknown / no-upstreams / row-newer-than-cutoff all keep the row,
  so a transient outage or in-flight download is never mistaken for genuine absence.

- **Bounded retry backoff** (`feat(upstream): backoff for transient retries`):
  `doRequest` retried transient transport errors (GOAWAY, http2 header timeout,
  conn reset, broken pipe) immediately. Add a capped exponential `waitRetryBackoff`
  (base `defaultRetryBackoff`, cap `defaultRetryBackoffCap`, override
  `Options.RetryBackoff`) that respects `ctx` cancellation. Genuine 404 still never
  retries (it returns a response, not a transport error).

- **Streaming fast-fail** (`fix(cache): fast-fail stalled progressive streaming`):
  `streamProgressiveChunks` only failed a stalled stream after the 30s per-chunk
  timeout. Add a stale-lock fast-fail (`time.Since(chunking_started_at) >
  cdcChunkingLockTTL` → not-found in ~200ms), never closing a truncated 200.

## Test plan

- [x] `task fmt` clean; changed packages lint clean
- [x] `task test` green; `pkg/cache/...` (cache + upstream) under `-race`
[...]

PR #1296, opened 2026-05-30
fix(cache): harden NAR cache-miss recovery & retry

kalbasit:
## Summary

Closes the remaining gaps behind the "phantom nar_file" defect — permanent
`error: file 'nar/<hash>.nar.xz' does not exist` 404s for paths whose narinfo
was just served. A backing-less `nar_file` row (`total_chunks=0`,
`chunking_started_at=NULL`) left by a failed background download was treated as
servable and short-circuited `GET /nar` to a terminal 404. Prior PRs
(#1255–#1290) fixed the core read path; this hardens the edges and prevents the
recurring regression.

- **Upstream retry (D5):** extend GET/HEAD retry beyond `GOAWAY` to also cover
  `http2: timeout awaiting response headers`, connection reset, and broken pipe
  via `isRetriableTransportError`. Genuine 404s return a response (not a
  transport error) and stay non-retryable.
- **Single source of truth (D1):** unify the duplicated, drift-prone
  servability checks in `GetNar` and `prePullNar`'s `hasAsset` behind one
  `isServable()` predicate (+ `narFileServable()`), so a placeholder row is
  never servable and the regression can't return via divergent checks.
  Behavior-preserving.
- **Recovery sweep:** `runCDCLazyRecovery` now skips backing-less stuck rows.
  `BackgroundMigrateNarToChunks` can only chunk a store-present NAR, so
  re-driving placeholder/genuinely-absent rows every interval was futile
  `error fetching nar from store: not found` spam that indefinitely retried
  unmigratable hashes. Such rows are recovered on demand by `GetNar`.

No schema migration. OpenSpec change `fix-phantom-nar-cache-miss` archived and
its delta specs synced into `openspec/specs/`.

## Test plan

- [x] `task fmt` clean; changed packages lint clean
- [x] `task test` green (all packages; `pkg/cache` + `pkg/cache/upstream` under `-race`)
- [x] `phantom_recovery_test.go` — reproduces narinfo 200 → failed download → recovery on next request, plus the genuine-404 path
- [x] `transient_retry_test.go` — new retriable errors retried; genuine 404 not retried
- [x] `recovery_skip_internal_test.go` — sweep skips backing-less rows, re-drives store-present ones (RED→GREEN verified)

PR #1298, opened 2026-05-30
fix(cache): address review feedback on #1296/#1297

kalbasit:
## Summary

Follow-up to #1297 (stacked on it) addressing review feedback from CodeRabbit
(#1296 outside-diff) and Gemini (#1297). Each fix is its own commit, TDD'd.

- **Persist the lazy-recovery cursor** (CodeRabbit, #1296): the keyset cursor lived
  only in the scheduled closure's memory, so a pod restart or `cdc-lazy-recovery`
  lock handoff reset the scan to 0 — a low-id backlog of backing-less placeholders
  could then refill every batch and starve higher-id re-drivable rows in
  clustered/restarting deployments. Persist it in the existing config KV table
  (`cdc_lazy_recovery_cursor`) via `loadRecoveryCursor`/`saveRecoveryCursor`. The
  distributed lock serializes sweeps, so a plain upsert suffices; **no migration**.

- **GC only when all linked narinfos are absent** (Gemini, #1297): the placeholder GC
  resolved a single narinfo via `First()`; multiple store paths can share one NAR, so
  a still-present narinfo could be missed and an active NAR wrongly deleted. Query
  **all** linked narinfos and delete only when every one is genuinely absent upstream.

- **Skip backoff on the final retry attempt** (Gemini, #1297): `waitRetryBackoff` ran
  after the last failed attempt too, padding the latency of an already-doomed request.
  Guard with `i < defaultHTTPRetries-1` (and flatten `doRequest` into guard clauses).

## Test plan

- [x] `task fmt` clean; changed packages lint clean
- [x] `pkg/cache/...` (cache + upstream) pass under `-race`; full unit suite green
- [x] Cursor: `TestRunCDCLazyRecoveryCursorPersists` (persists + resumes across a fresh closure)
- [x] GC: `TestRecoveryGCKeepsWhenAnyLinkedNarInfoPresent`
- [x] Backoff: `TestDoRequest_NoBackoffOnFinalAttempt`
- [ ] CI cohorts green

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -6576,6 +6576,13 @@ func (c *Cache) loadRecoveryCursor(ctx context.Context) int {
 		Where(entconfigentry.KeyEQ(cdcRecoveryCursorKey)).
 		Only(ctx)
 	if err != nil {
+		// A missing entry is expected (first run / after a wrap to 0). Any other
+		// error (transient DB issue, lock) resets the scan to 0, which can revive the
+		// low-id starvation this cursor prevents, so surface it for diagnosis.
+		if !database.IsNotFoundError(err) {
+			zerolog.Ctx(ctx).Warn().Err(err).Msg("failed to load CDC lazy-recovery cursor; resuming from 0")
+		}
+
 		return 0
 	}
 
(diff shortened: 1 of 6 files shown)


kalbasit:

- #1296
    - #1297
        - #1298 ◀

PR #1299, opened 2026-05-30
fix(cache): recover from ambiguous storage errors & bound chunk-wait

kalbasit:
## Summary

On a high-latency or shared-storage backend (e.g. the `local` backend on a network
filesystem with multiple replicas), two failure modes still surfaced to clients as
spurious cache-miss 404s and gateway 504s despite #1296/#1297/#1298:

1. **False-purge cascade.** `HasNar` collapsed *every* error into `false`, so an
   ambiguous/stale NFS stat was indistinguishable from a confirmed absence. The
   narinfo purge guard then destructively purged the narinfo, which clients saw as
   `does not exist in binary cache` (the `"requesting a purge"` log line).
2. **Unbounded per-chunk wait.** Progressive CDC streaming waited a hard-coded 30s
   per chunk; a legitimately slow chunk on high-latency storage accumulated past the
   reverse-proxy timeout and surfaced as a 504 / truncated transfer.

## Changes

- Add `storage.NarStore.StatNar` returning **present / confirmed-absent / unknown**;
  implemented in `local` (`os.IsNotExist`) and `s3` (`NoSuchKey`); `HasNar` now
  delegates to it.
- `getNarInfoFromDatabase` skips the purge on an **ambiguous (non-not-found)** storage
  result and keeps the narinfo, letting the next request re-evaluate.
- Make the per-chunk wait operator-configurable via `cache.cdc.chunk-wait-timeout`
  (`CACHE_CDC_CHUNK_WAIT_TIMEOUT`, default 30s), replacing the hard-coded value so it
  can be aligned with the gateway timeout.
- Storage-backend selection docs (README + `config.example.yaml`): `local` assumes
  single-writer POSIX semantics; multi-replica should use an object store; CDC suits
  low-latency storage.

## Notes

- Most of the original investigation's scope was **already shipped** in the deployed
  `sha-7d071cf` (phantom recovery, aborted-chunking guards); this PR targets the two
  remaining gaps that map to the observed production failures.
- **Deferred to a follow-up:** pre-commit chunk-set verification with
  fallback-to-redownload — reworking `GetNar` servability is regression-risky and
  per-chunk pre-flight stats add load to the same slow backend. Its spec requirement
  was removed before archive so the spec matches what is implemented.
- openspec change `fix-nar-serving-failures` archived (delta specs synced into
  `openspec/specs/`).

## Test plan

- [x] `task fmt`, `task lint`, `task test` green
- [x] TDD, RED verified: `TestGetNarInfoFromDatabase_AmbiguousStorageErrorDoesNotPurge`
      (RED = "the narinfo was purged"), `TestStreamProgressiveChunks_ChunkWaitTimeoutIsConfigurable`
- [ ] CI `nix flake check` (per-backend integration cohorts)

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,25 @@ Your cache will be available at `http://localhost:8501`. See the [Quick Start Gu
 
 See the [Deployment Guide](https://docs.ncps.dev/user-guide/deployment.html) for detailed setup instructions.
 
+### Choosing a storage backend
+
+The `local` filesystem backend assumes single-writer POSIX semantics: one process
+owning the directory, with read-after-write consistency. Use it for single-instance
+deployments.
+
+For **multiple replicas, use the S3-compatible backend.** Do not point the `local`
+backend at a shared network filesystem (NFS/SMB) written by more than one replica:
+network filesystems provide only close-to-open consistency with client-side
+attribute caching, so one replica can read stale metadata for a file another replica
+just wrote — which presents to ncps as a NAR that is in the database but "missing"
+from storage, and to clients as spurious cache misses or truncated transfers. An
+object store gives every replica a single, read-after-write-consistent view.
+
+Storage latency also drives the CDC decision (see `cache.cdc` in
+[`config.example.yaml`](config.example.yaml)): CDC's many small chunk reads suit
+low-latency backends, while high-latency or spinning-disk storage is better served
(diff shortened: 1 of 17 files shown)


kalbasit:

- #1299 ◀

PR #1300, opened 2026-05-30
feat(helm): expose cache.cdc.chunk-wait-timeout

kalbasit:
## Summary

The `cache.cdc.chunk-wait-timeout` flag (`CACHE_CDC_CHUNK_WAIT_TIMEOUT`, default `30s`) added in #1299 bounds the per-chunk wait during progressive CDC streaming so operators can align it with their reverse-proxy/gateway timeout on high-latency storage. The flag was registered in the binary but not reachable from the Helm chart.

- Add `config.cdc.chunkWaitTimeout` (default `null`) to `values.yaml` and render `chunk-wait-timeout` conditionally in the ConfigMap, mirroring the existing optional `cdc.*` knobs (quoted duration, omitted when unset so the binary's `30s` default applies).
- Add helm unit tests: key rendered when set, omitted when null.
- Document the knob in the Helm chart reference and example, and document the underlying `--cache-cdc-chunk-wait-timeout` flag in the CLI and CDC references (the flag shipped in #1299 without these doc updates).

No Go, database, or migration changes — this only surfaces an existing flag through the chart.

## Test plan

- [x] `task fmt` clean (0 changed)
- [x] `helm unittest charts/ncps` — 141/141 pass (RED→GREEN on the two new assertions)
- [x] CI `nix flake check` (helm unit tests)

### DIFF
--- a/charts/ncps/templates/configmap.yaml
+++ b/charts/ncps/templates/configmap.yaml
@@ -55,6 +55,9 @@ data:
         {{- if .Values.config.cdc.lazyCleanupSchedule }}
         lazy-cleanup-schedule: {{ .Values.config.cdc.lazyCleanupSchedule | quote }}
         {{- end }}
+        {{- if .Values.config.cdc.chunkWaitTimeout }}
+        chunk-wait-timeout: {{ .Values.config.cdc.chunkWaitTimeout | quote }}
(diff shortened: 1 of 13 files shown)


kalbasit:

- #1300 ◀

PR #1301, opened 2026-05-30
feat(cache): add migrate-chunks-to-nar (reverse CDC migration)

kalbasit:
## Summary

Adds `migrate-chunks-to-nar`, the reverse of `migrate-nar-to-chunks`: it reconstructs CDC-chunked NARs back into whole files so a deployment can **exit CDC**. Previously the forward migration was one-way (with `delete-delay` the whole files are GC'd), so once chunked there was no supported path back to whole-file storage.

### Cache — `Cache.MigrateChunksToNar(ctx, narURL, forceReclaim)`

- Per-hash `TryLock("migration-to-nar:"+hash)`, mirroring the forward migration.
- Reconstructs from chunks (`getNarFromChunks`), verifies the assembled SHA-256 against the linked narinfo's `NarHash` (+ size), and writes the whole file via `narStore.PutNar` (bypassing `c.PutNar` so CDC does not re-chunk it). **Verified-or-nothing** — a NAR is never de-chunked unless it verifies.
- **Write → flip → reclaim** ordering for crash-safe resumability.
- **Reclaim is deferred by default**: a serve that began streaming from chunks before the flip may still be reading those chunk files, so deleting them mid-stream would truncate that transfer (the failure class #1296/#1297 fixed). Orphaned chunks are left for the GC. `--force-reclaim` opts into immediate, dedup-safe reclamation for drained/maintenance-window runs.
- Sentinels: `ErrNarAlreadyWholeFile`, `ErrNoNarHashToVerify` (no `NarHash` → left chunked, not de-chunked unverified), `ErrNarHashMismatch`.

### CLI — `ncps migrate-chunks-to-nar`

Mirrors `migrate-nar-to-chunks` (storage/db/lock flags, `errgroup` concurrency, `--dry-run`). Adds `--force-reclaim`. Per-NAR failures are isolated (the batch continues) and the command exits non-zero if any NAR failed. Registered in the root command.

### Helm chart

Opt-in `migrate-chunks-to-nar` Job (`post-install`/`post-upgrade` hook) with `migrateChunksToNar.{enabled,dryRun,forceReclaim,concurrency}`, reusing the deployment's db/storage/redis config.

### Docs

New "Operations / Chunks to NAR Migration" page (CLI, `--dry-run`, `--force-reclaim` with the in-flight-serve safety note, the Helm Job, verification/skip behavior, failure isolation); linked from the Operations index and the CDC feature page; `--cache-cdc-chunk-wait-timeout` documented in the CDC flags table.

### openspec

New `chunks-to-nar-migration` capability (proposal/design/specs/tasks), archived and synced.

## Test plan

- [x] TDD, RED-verified. Cache tests under `-race`: reconstruct+verify+store, record flip, default-leaves-chunks-for-GC, force-reclaim-deletes-orphans, force-reclaim-retains-shared.
- [x] CLI tests: success (default-defer), force-reclaim, dry-run-makes-no-changes, hash-mismatch → non-zero exit without destroying data.
- [x] `task fmt` / `task lint` (0 issues) / `task test` green; 146 helm unittests pass.
- [ ] CI `nix flake check`


kalbasit:

- #1301 ◀

PR #1304, opened 2026-05-31
feat(cache): allow disabling CDC after migrate-chunks-to-nar

kalbasit:
## Summary

- Removes the hard startup error (`ErrCDCDisabledAfterEnabled`) that prevented operators from setting `cdc.enabled: false` on a cluster that previously had CDC enabled
- Replaces the guard with a `DeleteCDCConfig` call that clears all four stored CDC config keys (`cdc_enabled`, `cdc_min`, `cdc_avg`, `cdc_max`), so a future re-enable is treated as a fresh first boot
- After `ValidateOrStoreCDCConfig` succeeds with CDC disabled, `serve.go` queries the count of `nar_file` rows with `total_chunks > 0` and logs a structured `Warn` if any remain — pointing operators to `migrate-chunks-to-nar` as the remediation
- Any un-migrated chunked NARs become cache misses after CDC is disabled and are re-fetched from upstream on demand (handled by the existing cache-miss-recovery path)
- Adds an **Exiting CDC** section to the Chunks to NAR Migration doc with a step-by-step disable procedure

## Test plan

- [x] `task fmt` passes (no changes)
- [x] `task lint` passes for all changed packages
- [x] `task test` passes — all unit tests green
- [x] Existing `testValidateCDCDisableAfterEnabled` updated: asserts `nil` return, all four CDC config keys absent after disable, and re-enable succeeds as a fresh first boot

### DIFF
--- a/docs/docs/User Guide/Operations/Chunks to NAR Migration.md
+++ b/docs/docs/User Guide/Operations/Chunks to NAR Migration.md
@@ -103,6 +103,47 @@ migrateChunksToNar:
 
 `forceReclaim` maps to `--force-reclaim` and `dryRun` to `--dry-run`. The Job reuses the deployment's database, storage, and Redis configuration. Because it is gated by `migrateChunksToNar.enabled`, nothing is rendered when disabled (the default).
 
+## Exiting CDC
+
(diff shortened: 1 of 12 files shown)


kalbasit:

- #1304 ◀
    - #1305

PR #1305, opened 2026-05-31
feat(cache): implement CDC drain mode

kalbasit:
## Summary

Fixes a fundamental design flaw in PR #1304 (`allow-disabling-cdc`):
when CDC was disabled after having been enabled, chunked NARs became
cache misses and the stored DB config was cleared at startup, breaking
concurrent `migrate-chunks-to-nar` runs. Large deployments need a
graceful transition where existing chunks continue to serve while
migration runs at its own pace — mirroring the existing off→on behavior.

### Changes

**Cache layer — split write gate from read gate**
- Add `isChunkStoreAvailable() bool { return c.chunkStore != nil }` in
  `pkg/cache/cache.go`
- Change 4 read-path sites from `isCDCEnabled()` to
  `isChunkStoreAvailable()`: `isServable`, `HasNarInChunks`,
  `maybeCDCNormalizeNarInfoURL`, `getNarFromChunks`
- All ~22 write-path sites (`putNarWithCDC`, lazy-chunking, GC, etc.)
  remain gated on `isCDCEnabled()` — no writes in drain mode

**Config layer — preserve stored config during disable**
- Revert `validateCDCConfig`'s `DeleteCDCConfig` call from PR #1304;
  return `nil` and leave all four CDC keys intact so drain mode can
  re-initialize the chunk store on every restart

**Serve layer — drain mode with auto-complete detection**
- Extract `initCDCDrainMode` helper in `pkg/ncps/serve.go`
- Detect drain mode at startup: `storedWasEnabled=true && !cdcEnabled`
- If chunk count > 0: init chunk store read-only, log warning with count
- If chunk count = 0: call `DeleteCDCConfig`, log "drain complete",
  start fully disabled (self-completing transition, no operator command)

**migrate-chunks-to-nar — use chunk count not DB flag**
- Remove the `cdc_enabled=true` DB check
- Replace with `TotalChunksGT(0)` count; if 0 → "nothing to migrate"
  and exit 0; otherwise proceed regardless of DB enabled flag

### State machine

| DB `cdc_enabled` | config `cdc.enabled` | chunks | Mode |
|---|---|---|---|
| absent | false | — | fully disabled |
| true | true | — | full CDC (read + write) |
| true | false | yes | drain (reads from chunks, writes whole-file) |
| true | false | no | auto-clears config → fully disabled |

## Test plan

- [x] `task fmt` passes
- [x] `task lint` passes (removed 1 nestif issue vs baseline)
- [x] `task test` passes — all unit tests green
- [x] New test `testCDCDrainModeGetNarServesFromChunks`: end-to-end
  GetNar streaming in drain mode (cdcEnabled=false, chunkStore≠nil)
- [x] New test `testGetNarInfoDrainModeNormalizesChunkedNar`: GetNarInfo
  URL normalization in drain mode
- [x] New test `TestMigrateChunksToNar_CLI_NothingToMigrate`: command
  exits 0 when no chunked NARs exist regardless of DB state
- [x] Updated `testCDCDisabledWithChunkedNARsInDB`: reflects drain mode
  semantics (HasNarInChunks=true with chunk store set; false when nil)
- [x] Updated `testValidateCDCDisableAfterEnabled`: asserts all 4 DB
  keys still present after disable (opposite of PR #1304 assertion)

### DIFF
--- a/openspec/changes/archive/2026-05-30-cdc-drain-mode/.openspec.yaml
+++ b/openspec/changes/archive/2026-05-30-cdc-drain-mode/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-05-31
--- a/openspec/changes/archive/2026-05-30-cdc-drain-mode/design.md
+++ b/openspec/changes/archive/2026-05-30-cdc-drain-mode/design.md
@@ -1,0 +1,146 @@
+## Context
+
+`isCDCEnabled()` in `pkg/cache/cache.go` returns `c.cdcEnabled && c.chunkStore != nil`.
+It is used as a single gate for **both writes and reads**: whether to store a new NAR
+as chunks (write) and whether to serve an existing NAR from the chunk store (read).
+`SetChunkStore` is only called from `serve.go` when `cdcEnabled=true`, so setting
+`cdc.enabled: false` leaves `chunkStore == nil`, making every call to `isCDCEnabled()`
+return false — including the read paths that check for existing chunks.
+
+Affected read-path call sites:
+- `isServable` (line 3900): short-circuits to `false` when `!isCDCEnabled()`,
+  so chunked NARs are never considered servable → cache miss → upstream re-fetch.
+- `HasNarInChunks` (line 6994): same guard, returns `false` unconditionally.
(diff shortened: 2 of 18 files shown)


kalbasit:

- #1304
    - #1305 ◀

PR #1306, opened 2026-05-31
fix(helm): drop sync hook from migration jobs

kalbasit:
## Summary

- Remove `helm.sh/hook*` annotations from the `migrate-chunks-to-nar` Job so it becomes a regular release resource instead of an ArgoCD sync hook. Two OOMKills today left the hook in a failed state and blocked all subsequent ncps syncs until manually deleted.
- Add the missing `migrate-nar-to-chunks` Helm Job (forward CDC migration — whole NAR → chunks), which had no chart representation at all.
- Both Jobs are disabled by default (`enabled: false`) and auto-cleanup via `ttlSecondsAfterFinished: 3600`.

## Why not a hook

ArgoCD hooks are appropriate for short-lived, mandatory, per-deploy operations (e.g. DB schema migrations). The chunks migration is long-running (10+ min), one-time, and opt-in — hooking it gates every sync on its outcome and re-runs it on every deploy until it succeeds. The right pattern is: enable in values → commit → ArgoCD creates the Job → disable when done.

## Test plan

- [x] `helm unittest charts/ncps` — 151/151 tests pass
- [x] `task fmt` — exit 0
- [x] Hook annotation absent from rendered Job (`isNull` assertion added to test)
- [x] New `migrate-nar-to-chunks` Job renders correctly with `--dry-run` and `--concurrency` passthrough

### DIFF
--- a/charts/ncps/templates/migrate-nar-to-chunks-job.yaml
+++ b/charts/ncps/templates/migrate-nar-to-chunks-job.yaml
@@ -1,27 +1,27 @@
-{{- if .Values.migrateChunksToNar.enabled -}}
+{{- if .Values.migrateNarToChunks.enabled -}}
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: {{ include "ncps.fullname" . }}-migrate-chunks-to-nar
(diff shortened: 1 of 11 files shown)


kalbasit:

- #1306 ◀

PR #1307, opened 2026-05-31
feat(helm): configurable UID/GID via containerDefaults

kalbasit:
## Summary

- Remove all hardcoded and default security context values from `values.yaml` — `podSecurityContext`, `securityContext`, and all per-job `securityContext` blocks now default to `{}`
- Add `containerDefaults.securityContext` as a global fallback applied to every container via deep-merge (per-container values win)
- Add `initImage.securityContext` for the `create-db-dir` busybox init container, which previously hardcoded `runAsUser/runAsGroup: 1000` in four templates — overriding `podSecurityContext` and breaking operators who changed the pod identity
- Use `mergeOverwrite + deepCopy` rather than `merge` to correctly handle boolean `false` overrides (Go's zero value for bool causes `merge` to silently restore the global's `true`)
- Wrap all `podSecurityContext` rendering in `{{- with }}` so an empty map produces no key instead of `securityContext: {}`

**Breaking change:** Bare installations will no longer have any security context defaults applied. Operators who relied on the previous defaults should set the following in their values:

```yaml
podSecurityContext:
  runAsNonRoot: true
  runAsUser: 1000
  runAsGroup: 1000
  fsGroup: 1000
  fsGroupChangePolicy: OnRootMismatch
  seccompProfile:
    type: RuntimeDefault

containerDefaults:
  securityContext:
    allowPrivilegeEscalation: false
    capabilities:
      drop: [ALL]
    readOnlyRootFilesystem: true
    runAsNonRoot: true
    runAsUser: 1000
    runAsGroup: 1000
```

## Test plan

- [x] `helm unittest charts/ncps` — 155 tests pass
- [x] `task fmt` — clean exit
- [x] `task test` — all unit tests pass
- [x] New deployment tests: no-defaults, containerDefaults propagation, per-container override, initImage override, fallback


kalbasit:

- #1307 ◀

PR #1313, opened 2026-06-02
fix(storage/local): don't wipe store/tmp on startup

kalbasit:
## Summary

- Remove `os.RemoveAll(storeTMPPath())` from `setupDirs()` in `pkg/storage/local`
- On a shared NFS PVC with two replicas, the incoming pod was deleting
  in-flight downloads owned by the outgoing pod during rolling updates,
  racing with active writes and crashing on startup
- `os.MkdirAll` (called immediately after) already handles both the
  "dir exists" and "dir doesn't exist" cases idempotently — the `RemoveAll`
  was never necessary for correctness, only for a "clean slate" that is
  unsafe in a multi-pod topology
- Leftover partial files are benign: the write path uses atomic
  tmp-then-rename and partials are never promoted to the NAR path

## Test plan

- [x] `task fmt` — no changes
- [x] `task lint` — 0 issues
- [x] `task test` — all packages pass
- [x] `TestSetupDirsIdempotent` — calling `New()` twice on same dir succeeds
- [x] `TestSetupDirsPreservesExistingTmpFiles` — existing tmp files survive startup
- [x] `TestNew/store/tmp_files_are_preserved_on_boot` — updated to assert new behaviour

### DIFF
--- a/openspec/changes/archive/2026-06-01-fix-store-tmp-mkdir-idempotent/.openspec.yaml
+++ b/openspec/changes/archive/2026-06-01-fix-store-tmp-mkdir-idempotent/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-02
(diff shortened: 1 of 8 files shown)


kalbasit:

- #1313 ◀

PR #1317, opened 2026-06-03
fix(cdc): heal orphaned mid-chunking NAR state via migration-lock liveness

### LINKED ISSUE #1230: CDC: async chunking can leave orphan state if server is killed mid-chunking
## Summary

When CDC is enabled, `Cache.storeNarWithCDCFromReader` returns the response to the client (200 + NAR bytes) **before** the background chunker has finished writing all chunks. If the server is killed (or otherwise restarts) while a background chunker is still in flight, the database is left with:

- `nar_files.total_chunks = 0`
- `nar_files.chunking_started_at` SET (recent)
- Partial rows in `nar_file_chunks` for chunk indices `0..M` where `M < N` (the would-be total)
- The corresponding chunk blobs exist in storage for indices `0..M`

On the next restart, any client request for the same NAR enters `getNarFromChunks` (because the lock is "fresh" — well under `cdcChunkingLockTTL = 1h`), streams chunks `0..M`, then hits the empty-batch path of the wait loop in `pkg/cache/cache.go` (~line 6730 / 6912). It waits up to `maxWaitPerChunk = 30s` for chunk `M+1` to appear, then errors with:

> `timeout waiting for chunk N after 30s: context deadline exceeded`

The client (nix) sees a truncated NAR stream and fails with `error: failed to read compressed data (Truncated zstd input)` (or a similar truncation error for plain NAR).

## Affected configuration

- `--cache-cdc-enabled` (regardless of `--cache-cdc-lazy-chunking-enabled`)
- Reproduces on **all** database engines (SQLite, PostgreSQL, MySQL)
- Reproduces with **local** storage; **S3 does not** reproduce — S3 PUT latency naturally serializes the chunker enough that it usually finishes before SIGTERM

## Reproduction

This was found via an end-to-end migration test (`dev-scripts/test-migration-e2e.py` on branch `test-dev-scripts-add-e2e`), but the bug is independent of any migration. Minimal repro:

1. Start ncps with CDC enabled and local storage:
   ```
   python3 dev-scripts/run.py --db sqlite --storage local --enable-cdc
   ```
2. Have a client fetch a NAR with a substantial closure (e.g. `nix build --substituters http://127.0.0.1:8501 nixpkgs#hello`). Look for `download of nar complete (CDC chunking in background)` in the server log.
3. **Immediately** SIGTERM the server (don't wait for chunking to settle).
4. Inspect the DB — at least one `nar_files` row will have `total_chunks=0` and `chunking_started_at` set, and `nar_file_chunks` will have fewer rows than the NAR ought to be chunked into.
5. Restart the server.
6. Re-fetch the same NAR. After ~30s of partial streaming, the client errors with `Truncated zstd input` (or similar).

Empirical numbers from one repro on the glibc-2.40 NAR (30 MB uncompressed, expected ~304 chunks):

| | total_chunks | chunking_started_at | nar_file_chunks rows | chunks rows |
|---|---|---|---|---|
| Pre-shutdown (in progress) | 0 | set | ~266 partial | shared |
| After failed re-serve | 0 | reset to fresh restart time | 266 (same partial set) | unchanged |

Server log on the failed second serve:
```
error: timeout waiting for chunk 266 after 30s: context deadline exceeded
status:200 elapsed:30175 bytes:8330215 (out of 30180096 expected)
```

## Why the 30s wait can never succeed

`maxWaitPerChunk` (`pkg/cache/cache.go:6702`) is **30s**.
`cdcChunkingLockTTL` (`pkg/cache/cache.go:2142`) is **1 hour**.

After a crash mid-chunking, `chunking_started_at` is well under the 1h lock TTL, so the read path treats the lock as live and waits for chunks that no one is producing. The 30s read timeout fires long before the 1h stale-lock recovery would clear the orphan state.

## Suggested fix (preferred: startup recovery)

On server start, scan `nar_files` for rows matching `total_chunks = 0 AND chunking_started_at IS NOT NULL`, treat them as orphaned, delete their `nar_file_chunks` rows + nullify both fields, and let the next request re-download cleanly. Orphan chunks in storage can be reaped by the existing chunk GC.

Alternatives that were considered:

1. **Block on chunking before responding.** Safest but introduces client-perceived latency proportional to NAR size; defeats the progressive-streaming optimization.
2. **Graceful-shutdown handshake.** On SIGTERM, wait up to N seconds for in-flight chunkers. Still leaves a window for SIGKILL / OOM-killer crashes.
3. **Startup recovery.** ← **preferred.** Self-heals from any unclean shutdown without changing the hot path.
4. **Lower `cdcChunkingLockTTL` below `maxWaitPerChunk`.** Trades the bug for a lot of false-positive "stale lock" recoveries during legitimate slow chunking.

## Scope

This is **not** a migration bug — the dbmate→Ent migration (PR #1229 + follow-ups) faithfully preserves whatever state the previous server left in the schema. The e2e migration test just happens to exercise the exact "kill server during async work" pattern that triggers this. CDC+local fails after migration; CDC+S3 passes — same migration code path, different storage timing.

Found via the `test-dev-scripts-add-e2e` branch in the same stack as PR #1229.

mvanhorn:
## Summary

Fixes the orphaned mid-chunking state described in #1230. When CDC is enabled, the
background chunker runs after the client is already served; a SIGKILL/OOM mid-chunking
leaves a `nar_file` row with `total_chunks = 0`, a **fresh** `chunking_started_at`, and a
partial set of `nar_file_chunks`. The read path treated the fresh lock as a live chunk,
streamed chunks `0..M`, then stalled `maxWaitPerChunk` (30s) on a chunk no producer would
write — the client saw `Truncated zstd input`.

The fix uses the **per-hash migration lock** (`migrationLockKey(hash)`, held by the
download-path chunker for the whole operation via `withNarMigrationLock`) as the single
cross-instance liveness signal, replacing the 1h `cdcChunkingLockTTL` *age* — a poor
liveness proxy — in the three places it was used:

1. **Read path (the actual symptom fix).** `isServable` probes the lock (new
   `cdcChunkerLive`, fail-safe to "live") for the ambiguous `total_chunks == 0 &&
   fresh-lock` case. A **free** lock means the producer is dead → the row is not servable →
   `GetNar` re-downloads cleanly **before** any bytes are streamed. A **held** lock keeps
   the row servable so legitimate progressive chunking still streams.
2. **Write path.** `findOrCreateNarFileForCDC` always runs while holding the migration
   lock, so a fresh in-progress row it encounters belongs to a dead chunker. It now
   reclaims the partial chunks and re-chunks regardless of age, instead of refusing a fresh
   row with `ErrAlreadyExists` (which would strand the orphan and block the re-download).
   The redundant in-transaction age gate in `clearStaleCDCChunkingLockWithEntTx` is removed.
3. **Cron.** `runCDCLazyRecovery` reaps in-progress orphans of **any age**, gated solely on
   a per-row `TryLock(migrationLockKey(hash))` (skip if a peer holds it, reap if free); the
   1h `staleCutoffTime` predicate is gone. This handles rows that are never re-requested and
   reclaims their orphaned chunk blobs.

## Why this is HA-safe

In a multi-replica shared-DB deployment, a `total_chunks = 0` in-progress row may belong to
a peer's **live** chunker — `total_chunks` stays 0 for the whole healthy operation, so age
can't distinguish live from dead. The migration lock can: a live chunker (this instance or
a peer) holds it, so every reaping/serving/takeover decision keys off the lock, never off
age. The lazy-recovery cron is additionally serialized across instances by
`withTryLock("cdc-lazy-recovery")`.

## Also addressed (review threads on this PR)

- Background CDC chunking that returns `ErrMigrationInProgress` (a peer holds the lock) is a
  benign debug no-op via `reportBackgroundCDCError` — no error-level log, no
  `downloadState.setError` — instead of HA log spam / spurious alerts.
- The lazy-chunking-disabled skip uses a distinct `lazy_chunking_disabled_skip_count`
  counter instead of inflating `stale_recovery_skip_count`.

## Testing

- New internal tests (TDD): read-path liveness gate (`read_path_liveness_internal_test.go`),
  write-path orphan takeover (`cdc_chunker_takeover_internal_test.go`), benign
  `ErrMigrationInProgress` handling (`background_cdc_error_internal_test.go`), cron
  any-age reaping + lock-held skip + counter disambiguation
  (`orphan_lazy_recovery_internal_test.go`, `recovery_skip_counter_internal_test.go`).
- The two progressive-streaming tests now hold the migration lock (via an `export_test.go`
  seam) to faithfully simulate a live chunker.
- `task fmt`, `task lint` (0 issues), and `task test` all pass.

## Known residual (accepted)

If a chunker dies *after* `isServable` passed (lock held) but mid-stream, the wait loop
still only fast-fails on >1h staleness — but bytes are already committed, so a clean
response isn't possible regardless. The dominant restart repro from #1230 is fully fixed.

A spec for this behavior is captured under OpenSpec (`cdc-orphan-chunking-recovery`,
modifying the `cdc-chunking` capability).

Fixes #1230


pkg/ncps/serve.go
```diff
@@ -1182,6 +1182,10 @@ func createCache(
 		}
 
 		c.SetChunkStore(chunkStore)
+
+		if err := c.RecoverOrphanedChunkingOnStartup(ctx); err != nil {
```
kalbasit: **Blocking the HTTP listener on the startup scan.** `RecoverOrphanedChunkingOnStartup` runs synchronously here inside `createCache`, which returns at `serve.go:660` — before the listener opens at `server.ListenAndServe()` (`serve.go:698`). So the server accepts no connections until this scan finishes.

The scan isn't cheap: it walks every `total_chunks=0 AND chunking_started_at IS NOT NULL` row (batches of 1000), opens **one transaction per row**, and per batch calls `cleanupStaleLockChunks`, which issues a physical `cs.DeleteChunk` per orphaned chunk (`cache.go:2447`). On the local backend over NFS each of those is a network round-trip, and a single mid-chunking large NAR can be hundreds of chunks. A backlog delays readiness, and under a k8s rolling deploy can trip the new pod's readiness/liveness probes.

The natural home for this is the existing `runCDCLazyRecovery` cron (`cache.go:6521`), which deliberately keeps recovery off the boot path and is gated by `withTryLock`. Since this scan is already best-effort / log-and-continue, run it off the listener path rather than gating boot on storage-delete latency. (See the related comment on the recovery method itself for why the lazy mechanism is the right *shape* but not a drop-in fix.)

pkg/cache/cache.go
```diff
@@ -2224,6 +2224,142 @@ func (c *Cache) storeNarWithCDCFromReader(
 // it is considered stale and a new attempt may clean up the partial state and restart.
 const cdcChunkingLockTTL = time.Hour
 
+// RecoverOrphanedChunkingOnStartup heals CDC nar_file rows left mid-chunking
+// by a previous process crash. At process startup there is no live in-process
+// chunker, so any total_chunks=0 row with chunking_started_at set is orphaned.
+func (c *Cache) RecoverOrphanedChunkingOnStartup(ctx context.Context) error {
```
kalbasit: **The load-bearing premise here only holds single-instance — and the existing HA machinery doesn't close the gap.** The doc comment asserts *"At process startup there is no live in-process chunker, so any total_chunks=0 row with chunking_started_at set is orphaned."* ncps supports multi-replica shared-DB deployments, where that exact row is a peer's **live** in-flight chunk. The in-transaction recheck doesn't help (coderabbit flagged the predicate): `total_chunks` is still 0 mid-chunk, so the recheck passes and this reaps the peer's partial `nar_file_chunks` + clears its lock. Concrete failure: during a rolling deploy the new pod scans and deletes the in-flight state of the old pod still serving — manufacturing the very `Truncated zstd input` this PR set out to fix, on healthy traffic.

It's tempting to lean on the lazy-recovery / locking machinery to make this HA-safe. I traced it — worth being precise about what carries over and what doesn't:

1. **Shape: yes.** `runCDCLazyRecovery` is a cron gated by `withTryLock("cdc-lazy-recovery")`, so it never blocks boot and only one instance sweeps. Moving this recovery into that sweep fixes the boot-blocking concern (other comment) and simultaneous multi-instance sweeps. Do this regardless.

2. **Coverage / safety: no.** Lazy recovery only targets `chunking_started_at IS NULL` placeholders and re-drives them via `BackgroundMigrateNarToChunks`, which requires a whole-file NAR in the store (`HasNarInStore`, `cache.go:6621`) and holds `migrationLockKey(hash)` for its duration (`cache.go:7596`). A crash-orphaned CDC row has only partial chunks (no whole file), so that path would GC/skip it. More fundamentally, the download-path chunker that *creates* these rows — `storeNarWithCDCFromReader`, spawned at `cache.go:2806` — holds **no distributed lock**; its liveness is an in-process `upstreamJobs` entry (`cache.go:2812`) plus `chunking_started_at`. So a recovery sweep on another instance has no per-NAR lock to `TryLock` against to prove a row is genuinely dead vs. a peer mid-chunk. That's exactly why the read path + 1h `cdcChunkingLockTTL` backstop (`cache.go:7412`, `narFileServable` at `3900`) only treat a lock as stale after 1h — and why an age-threshold gate doesn't help either (a safe cutoff ≥ 1h is what the existing backstop already does, defeating the point of healing *fresh* orphans).

The gap is a missing cross-instance liveness signal, not age. Options: (a) scope fresh-reap to single-instance behind a config/advisory "sole writer" guard and document it; or (b) add a real per-NAR liveness lock/heartbeat for download-path chunking that recovery can consult. Either way the premise in this comment should be corrected — it's the assumption the whole function rests on.

kalbasit (dismissed):
Thank you for your contribution @mvanhorn, much appreciated!

I reviewed the PR and I have two concerns:

a) **This blocks boot.** Because the recovery scan runs synchronously before the HTTP listener starts, an instance with many or large orphaned NARs to recover can be slow to become ready — on Kubernetes this risks tripping the readiness/liveness probes and sending the pod into a crash loop.

b) **It's unsafe in HA deployments.** With multiple replicas sharing a database, a booting instance would interfere with NARs that another instance is actively chunking — deleting their partial chunks and clearing the lock, which can manufacture the very truncation error this PR is trying to fix.

I asked Opus to review the PR against the current codebase and validate these concerns; I've left its findings in the comments below.

I think this should instead be implemented in a way that's adjacent to (and shares code with) the lazy-CDC recovery path, using the same locking so that one instance can't step on another's in-flight work or block boot.

mvanhorn:
Reworked per your review. The synchronous startup scan is gone entirely - boot no longer touches recovery. Orphaned mid-chunking state is now recovered inside the existing CDC lazy-recovery cron job (shared code path, same per-NAR migration locking), so a booting instance can't step on another replica's in-flight chunking and the recovery cadence follows --cache-cdc-lazy-recovery-schedule. The recovery job now arms for both CDC modes since orphans can exist in either. Full pkg/cache and pkg/ncps test suites pass locally, including the pre-existing lazy-recovery contract tests.

pkg/cache/cache.go
```diff
@@ -2838,7 +2927,13 @@ func (c *Cache) pullNarIntoStore(
 				ds.storedOnce.Do(func() { close(ds.stored) })
 			}
 
-			cdcErr := c.storeNarWithCDCFromReader(ctx, cdcReader, narInfo.NarSize, &narURLForCDC, onNarFileReady)
+			cdcErr := c.storeNarWithCDCFromReaderWithMigrationLock(
```
kalbasit, line 3033: The background CDC chunking now runs under the per-hash migration lock. When a peer holds it (another replica, or a concurrent `MigrateNarToChunks`/`MigrateChunksToNar`/stale-recovery for the same hash), `withNarMigrationLock` returns `ErrMigrationInProgress`, and this block logs it at **Error** level (`CDC chunking failed in background…`) and calls `ds.setError`. In a multi-instance fleet this benign "someone else owns it" outcome will produce error-level log spam and spurious alerts — even though the in-flight client already streamed the bytes and the lock holder will persist the NAR. Suggest special-casing `errors.Is(cdcErr, ErrMigrationInProgress)` as a Debug/Info no-op (return without `setError`). The same applies to the simple-path copy around line 3039.

pkg/cache/cache.go
```diff
@@ -6624,6 +6767,12 @@ func (c *Cache) runCDCLazyRecovery(ctx context.Context, schedule cron.Schedule,
 					continue
 				}
 
+				if !c.GetCDCLazyChunkingEnabled() {
+					staleRecoverySkipCount++
```
kalbasit: This is the **lazy-stuck** branch (`ChunkingStartedAt == nil`), but skipping it here increments `staleRecoverySkipCount`, which is emitted as the log field `stale_recovery_skip_count` — a counter whose name implies a *stale in-progress lock* was skipped. When lazy chunking is disabled, every lazy candidate inflates the stale-recovery skip metric, making the observability misleading (and conflating it with the legitimate stale-skip at line 6730). Use a separate counter (e.g. `lazyChunkingDisabledSkipCount`) or don't fold it into the stale metric.

kalbasit (commented):
Two things on top of the inline comments:

**Title/description** — please update them now to match what actually landed: cron-based, TTL-gated stale-lock recovery plus a per-hash migration-lock refactor — not the startup scan (`RecoverOrphanedChunkingOnStartup`, fresh-row reaping, `serve.go` wiring) the current description describes. Worth fixing before merge so the squash commit message is accurate.

**PutNar upload path** — `putNarWithCDC` (cache.go:1658) now goes through the lock-wrapping `storeNarWithCDC`, but its error handling only special-cases `storage.ErrAlreadyExists`; `ErrMigrationInProgress` falls through as a hard error. So a `nix copy … /upload` of hash X that collides with a concurrent background migration/recovery of the same hash now fails where it previously succeeded — a plausible regression given the recent `…/upload` abort history. Recommend treating `errors.Is(err, ErrMigrationInProgress)` the same as `ErrAlreadyExists` here.

kalbasit:
Since the PR has been idle, I pushed `83b3245` to address the outstanding review feedback and complete the fix for #1230.

### What was open vs. now
- **A — boot-blocking startup scan** & **E — `ErrMigrationInProgress` hard `GetNar` failure**: already resolved earlier on the branch (`430bc18` moved recovery into the lazy-recovery cron; `coordinateDownload` polls/takes over).
- **B — multi-instance premise / age gate**, **C — `ErrMigrationInProgress` log spam**, **D — skip-counter conflation**: addressed in this push.

### Key finding from the review
The cron-only recovery did **not** actually fix the user-visible symptom. The issue's own repro (restart → immediately re-fetch) lands before the first cron run, and the read path still treated a **fresh** dead-chunker orphan as servable on the 1h `cdcChunkingLockTTL` — so it streamed `200` + chunks `0..M` then stalled `maxWaitPerChunk` on a chunk no producer would write (`Truncated zstd input`).

The fix is to use the **per-hash migration lock** (`migrationLockKey`, held by the download-path chunker for the whole operation via `withNarMigrationLock`) as the single cross-instance liveness signal, replacing the 1h age — which was a poor liveness proxy — in all three places it was used:

1. **Read path (the symptom fix)** — `isServable` probes the lock (new `cdcChunkerLive`, fail-safe to "live") for the ambiguous `total_chunks==0 && fresh-lock` case. A free lock ⇒ dead chunker ⇒ not servable ⇒ `GetNar` re-downloads cleanly **before** any bytes are streamed. A held lock keeps the row servable so legitimate progressive chunking still streams.
2. **Write path** — `findOrCreateNarFileForCDC` always runs while holding the migration lock, so a fresh in-progress row it sees is a dead orphan; it now reclaims and re-chunks regardless of age instead of refusing with `ErrAlreadyExists` (which would strand the orphan and block the re-download). The redundant in-transaction age gate in `clearStaleCDCChunkingLockWithEntTx` was removed.
3. **Cron** — `runCDCLazyRecovery` reaps in-progress orphans of any age, gated solely on the per-row `TryLock` (skip if a peer holds it, reap if free); the `staleCutoffTime` predicate is gone. This still handles rows that are never re-requested and reclaims their chunk blobs.

Plus the two smaller threads: benign `ErrMigrationInProgress` background-chunk failures are now a debug no-op (`reportBackgroundCDCError`, no error log / no `setError`), and lazy-chunking-disabled skips use a distinct `lazy_chunking_disabled_skip_count` instead of inflating `stale_recovery_skip_count`.

### Notes
- New tests are TDD'd (read-path liveness, write-path takeover, benign-error handling, counter). The two progressive-streaming tests now hold the migration lock to faithfully simulate a live chunker.
- `task fmt` / `task lint` (0 issues) / `task test` all pass.
- **Known residual (accepted):** if a chunker dies *after* `isServable` passed but mid-stream, the wait loop still only fast-fails on >1h staleness — but bytes are already committed so a clean response isn't possible anyway. The dominant restart repro is fully fixed.
- A spec was written for this under OpenSpec (`cdc-orphan-chunking-recovery`, modifying `cdc-chunking`).

kalbasit:
Thank you! @mvanhorn

mvanhorn:
Much appreciated, @kalbasit. Healing orphaned mid-chunking NAR state via the migration lock closes that crash-window corruption.

PR #1318, opened 2026-06-03
fix(cache): never surface purged narinfo as HTTP 500

kalbasit:
## Summary

A narinfo present in the database but missing its backing NAR is purged by `getNarInfoFromDatabase`, which signals the internal `errNarInfoPurged` sentinel. After an upstream re-pull, `GetNarInfo` re-read that sentinel verbatim and the narinfo HTTP handler mapped it to **HTTP 500** (`the narinfo was purged`). Nix treats a 500 as a hard error and aborts the build instead of falling back to its next substituter — `fsck` reported **3013** such backing-less narinfos, each a permanent 500.

A purge is internal cache maintenance, not a server fault, so it must never reach the client as a 500.

### Changes
- **`pkg/cache/cache.go`** — `GetNarInfo`'s post-pull re-read now maps `errNarInfoPurged` → `storage.ErrNotFound` (HTTP 404 → Nix upstream fallback). The sentinel is exported as `cache.ErrNarInfoPurged`.
- **`pkg/server/server.go`** — the narinfo `GET` error→status decision is extracted into `narInfoErrorStatus`, mapping the sentinel (and `ErrNotFound`) to **404** while writing only the generic status text, never the internal message.
- Deterministic cache and server tests, using an unexported context-scoped test seam (`withNarPrefetchDisabled`) to reproduce the otherwise timing-racy bug.

Scoped to serving semantics only. Non-goals (deferred): the NAR background-download race under distributed-lock contention (root cause of missing NARs) and reconciliation of the existing orphaned-narinfo backlog.

OpenSpec capability: `narinfo-purge-serving`.

## Test plan

- [x] `task fmt` clean
- [x] `task lint` — 0 issues
- [x] `task test` — all packages pass with `-race`
- [ ] Verify `ncps.nasreddine.com` returns 404 (not 500) for an orphaned narinfo hash after deploy

### DIFF
--- a/openspec/changes/archive/2026-06-03-fix-purged-narinfo-500/.openspec.yaml
+++ b/openspec/changes/archive/2026-06-03-fix-purged-narinfo-500/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-03
--- a/openspec/changes/archive/2026-06-03-fix-purged-narinfo-500/design.md
+++ b/openspec/changes/archive/2026-06-03-fix-purged-narinfo-500/design.md
@@ -1,0 +1,109 @@
+## Context
+
+`GetNarInfo` (`pkg/cache/cache.go:3302`) resolves a narinfo through three stages:
+
+1. **DB lookup** — `getNarInfoFromDatabase` (line 3330). Its purge guard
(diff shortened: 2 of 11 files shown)


kalbasit:

- #1318 ◀

PR #1319, opened 2026-06-03
fix(cache): 404 un-reassemblable chunked NARs instead of truncated 200

kalbasit:
## Summary

Production Nix clients were getting **truncated NAR downloads** from ncps
(`unexpected end of nar`, `truncated input`, `SSL_read: SSL_ERROR_SYSCALL`) on
responses that started as `HTTP 200`. Root cause, confirmed against the live DB:

- `nar_file_chunks.chunk_id -> chunks(id)` is `ON DELETE CASCADE`. Deleting a
  shared `chunks` row strips its junction links from **completed** chunked NARs
  (`total_chunks > 0`) **without** resetting `total_chunks`. Nothing enforced
  `count(links) == total_chunks`.
- The serve fast path only detected the gap (`expected N chunks but got M`)
  **inside the streaming goroutine**, after `server.go` had already written
  `200 OK` + `Content-Length` — so the client got a body it couldn't fulfill and
  Nix could not fall back.
- This is **distinct from #1230/#1317** (those cover the `total_chunks = 0`
  mid-chunking orphans). Of 2499 completed chunked NARs in prod, 71 had
  `links < total_chunks`.

### Fix

`getNarFromChunks` now counts junction links in its init transaction on the
`total_chunks > 0` fast path and returns `storage.ErrNotFound` **synchronously**
when `links != total_chunks`, before the pipe/response are committed → handler
responds `404` → Nix refetches upstream. `total_chunks` is the completion latch
(written only after all links commit), so this is never a mid-chunking race; the
progressive `total_chunks = 0` path is left untouched and stays HA-safe.

The `migrate-chunks-to-nar` driver already maps `ErrMissingChunk` to
`PurgeChunkedNar` + continue + report; the guard routes link-loss into that path,
so the 71 broken NARs self-heal (purge → refetch) on the next drain.

Specs synced: new `chunked-nar-serving-integrity` capability + 2 added
requirements on `cdc-drain-mode`.

## Test plan

- [x] `task fmt` / `task lint` (0 issues) / `task test` (all packages ok)
- [x] `go test -race ./pkg/cache ./pkg/server` — ok
- [x] New tests: serving 404-not-truncated-200, HA-safety on progressive path,
      missing-link → `ErrMissingChunk`, purge → clean cache-miss, and a CLI
      mix-migration test (skip broken + migrate good + report by hash)

### DIFF
--- a/openspec/changes/archive/2026-06-03-fix-truncated-chunked-nar-serving/.openspec.yaml
+++ b/openspec/changes/archive/2026-06-03-fix-truncated-chunked-nar-serving/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-03
--- a/openspec/changes/archive/2026-06-03-fix-truncated-chunked-nar-serving/design.md
+++ b/openspec/changes/archive/2026-06-03-fix-truncated-chunked-nar-serving/design.md
@@ -1,0 +1,138 @@
+## Context
+
+Nix clients receive truncated NAR downloads from production ncps: responses that
+begin as `HTTP 200` (with a `Content-Length`) then die mid-body
+(`unexpected end of nar`, `truncated input`, `SSL_ERROR_SYSCALL`). See
+`proposal.md` for the full motivation and production evidence.
+
+Current serve path (`pkg/cache/cache.go`):
+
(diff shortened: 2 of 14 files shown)


kalbasit:

- #1319 ◀

PR #1321, opened 2026-06-03
fix(cache): no purge on upload-only narinfo reads

kalbasit:
## Summary

`nix copy --to https://<host>/upload <closure>` aborted with
`cannot add 'X' to the binary cache because the reference 'Y' does not exist`
whenever the closure contained a **phantom narinfo** (DB row present, backing
NAR bytes absent).

The read-path purge guard destructively purged such narinfos on **every** read,
including on the `/upload` route. That made the destination's path-validity
answer **non-monotonic** within a single `nix copy`: a path was present at the
client's `queryValidPaths` phase, then purged before its reference was verified,
so Lix's `BinaryCacheStore::addToStore` reference check aborted the whole copy.
Each retry re-`PUT` the narinfo, which re-seeded the phantom and was purged again
— observed churn of 244 purge events in one production run.

## Fix

`getNarInfoFromDatabase` now skips `purgeNarInfo` when the request is upload-only
(`cache.IsUploadOnly`), returning the existing `ErrNarInfoPurged` sentinel that
`GetNarInfo` already maps to `storage.ErrNotFound` (HTTP 404) on that path. The
client receives 404 and re-uploads; the `PUT` overwrites the stale record.
Root/substituter purge-and-re-fetch behavior is unchanged.

Scope is the read path only — phantom *seed* prevention at write time
(`PutNarInfo` creating a `nar_file` record without bytes) is deliberately a
separate follow-up.

## Test plan

- [x] `task fmt` / `task lint` / `task test` pass (race detector on)
- [x] New `pkg/cache/upload_only_purge_internal_test.go`:
  - upload-only read → 404, narinfo row **not** purged
  - repeated upload-only reads → deterministic 404, non-mutating (monotonic)
  - non-upload-only read → still purges (regression boundary)
  - post-404 re-upload repairs the record (served `200`)
- [x] `narinfo-purge-serving` spec synced (added upload-only requirement; scoped the upstream-refetch requirement to non-upload-only)

### DIFF
--- a/openspec/changes/archive/2026-06-03-fix-upload-read-no-purge/.openspec.yaml
+++ b/openspec/changes/archive/2026-06-03-fix-upload-read-no-purge/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-03
--- a/openspec/changes/archive/2026-06-03-fix-upload-read-no-purge/design.md
+++ b/openspec/changes/archive/2026-06-03-fix-upload-read-no-purge/design.md
@@ -1,0 +1,96 @@
+## Context
+
+`GetNarInfo` (pkg/cache/cache.go) has a "purge guard": when a narinfo row exists
+but its backing NAR is absent from storage and no download is in flight, it calls
+`purgeNarInfo` (cache.go:~4357) — a destructive DB delete of the narinfo +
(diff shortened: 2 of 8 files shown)


kalbasit:

- #1321 ◀

PR #1323, opened 2026-06-04
fix(cache): NAR HEAD must verify bytes, not DB record

kalbasit:
## Summary

`nix copy --to https://<host>/upload <closure>` aborted with
`cannot add 'X' to the binary cache because the reference 'Y' does not exist`
even after the purge fix (#1321) and **even at a single replica** (so this is
not a cross-replica/NFS issue).

Root cause: the NAR `HEAD` handler (`getNar(false)`) returned `200` whenever
`GetNarFileSize` reported `size > 0`, and `GetNarFileSize` reads only the
`nar_file` **DB record** — it never checks that the NAR bytes exist. The
reference-check `GET` path, by contrast, decides existence from **actual bytes**
(`statNarInStore`). When a `nar_file` row exists without backing bytes (a
phantom), nix's pre-upload `HEAD /upload/nar/<h>.nar.zst` returns `200`, so nix
**skips uploading the NAR** (it uploads NAR-first, narinfo-last) and writes only
the narinfo → phantom → a dependent path's reference-verification `GET` `404`s →
Lix aborts. The mass `PutNar ... context canceled` errors are the *cascade* of
nix tearing down in-flight uploads at abort, not the cause.

## Fix

- Add `Cache.IsNarServable(ctx, narURL) (bool, error)` — tri-state servability
  from actual bytes (whole-file or chunks): `(true,nil)` present, `(false,nil)`
  confirmed absent, `(false,err)` ambiguous storage probe.
- Gate the HEAD size-shortcut on it: emit `200` only when the NAR is actually
  servable; otherwise fall through to `GetNar`, which resolves upload-only to
  `404` (so the client re-uploads) and the substituter path to upstream recovery
  — consistent with `GET`.

Scope is the HEAD existence check only. Write-path seed prevention
(`storeInDatabase` creating byteless `nar_file` rows) is a deliberate follow-up;
the HEAD fix alone breaks the failure chain. Follow-up to #1321.

## Test plan

- [x] `task fmt` / `task lint` / `task test` pass (race detector on)
- [x] `TestGetNar_HeadBytelessNarIs404` — byteless NAR HEAD → `404`, then `200`
  after the NAR is uploaded (fast path preserved). (Repurposed from the old
  `TestGetNar_HeadOptimization`, which had codified the buggy `200`.)
- [x] `TestGetNar_HeadAmbiguousStorageIsNotFalse200` — ambiguous storage probe
  must not yield a false `200`.
- [x] `IsNarServable` cache tests: whole-file present, byteless `(false,nil)`,
  ambiguous `(false,err)`.
- [x] `nar-cache-miss-recovery` spec synced (HEAD/servability requirement added).

### DIFF
--- a/openspec/changes/archive/2026-06-03-fix-nar-head-verify-bytes/.openspec.yaml
+++ b/openspec/changes/archive/2026-06-03-fix-nar-head-verify-bytes/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-04
--- a/openspec/changes/archive/2026-06-03-fix-nar-head-verify-bytes/design.md
+++ b/openspec/changes/archive/2026-06-03-fix-nar-head-verify-bytes/design.md
@@ -1,0 +1,111 @@
+## Context
+
+The NAR `HEAD` handler `getNar(false)` (pkg/server/server.go) has a fast-path
+optimization: it calls `c.GetNarFileSize(nu)` and, if that returns `size > 0`,
+writes `200` with `Content-Length` and returns — **without** confirming the NAR
+bytes exist. `GetNarFileSize` (pkg/cache/cache.go:1430) reads the `nar_file` row
+via `getNarFileFromDB`; it is a pure DB lookup.
+
+Every other path determines NAR existence from actual bytes:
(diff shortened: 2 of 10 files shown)


kalbasit:

- #1323 ◀

PR #1324, opened 2026-06-04
fix(cache): fall back to chunks when migration races serve

kalbasit:
## Summary

Fixes the flaky `TestCDCBackends/.../Mixed_Mode` failure seen in CI
(e.g. on #1322) — `error fetching the nar from the store: not found`.

The root cause is a TOCTOU race in the NAR read path. When a whole-file
NAR is served while CDC is enabled, `GetNar` triggers a **background**
NAR→chunks migration that deletes the whole file, then *synchronously*
serves from that same store. If the delete wins the race after
`serveNarFromStorageViaPipe` chose the store branch (`total_chunks`
still observed as `0`) but before `getNarFromStore` opens the file, the
read 404s — even though the NAR is fully available as chunks. In
production this surfaces as spurious 404s on freshly-migrated NARs.

Because the migration commits the chunks **before** deleting the whole
file, the NAR is always retrievable from chunks at that point. The fix:
when a store read returns `ErrNotFound` while CDC is enabled and the
request is uncompressed, fall back to `getNarFromChunks`.

- The chunk result is adopted **only on success**; if chunks are also
  absent, the original store `ErrNotFound` is preserved (genuine
  not-found semantics unchanged).
- Compressed requests are excluded — chunks are stored uncompressed, so
  a `.nar.xz` request keeps returning `ErrNotFound` and falls back
  upstream.

## Test plan

- [x] New deterministic regression test reproduces the race via a
  `narStore` wrapper (RED → GREEN; RED reproduced the exact CI error)
- [x] Added coverage for the compressed-request and genuinely-absent
  edges; documented the race in `testCDCMixedMode`
- [x] `TestCDCBackends` + new tests pass under `-race` (repeated runs)
- [x] `task fmt`, `task lint` (0 issues), `task test`, and full
  integration suite (`task test:auto`, real Postgres/MySQL/Redis) pass

## Spec

Adds the `cdc-chunking` requirement "Serving a whole-file NAR MUST be
resilient to a concurrent background migration" (4 scenarios), and
archives the OpenSpec change under
`openspec/changes/archive/2026-06-03-fix-flaky-cdc-mixed-mode-test/`.

### DIFF
--- a/openspec/changes/archive/2026-06-03-fix-flaky-cdc-mixed-mode-test/.openspec.yaml
+++ b/openspec/changes/archive/2026-06-03-fix-flaky-cdc-mixed-mode-test/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-04
--- a/openspec/changes/archive/2026-06-03-fix-flaky-cdc-mixed-mode-test/design.md
+++ b/openspec/changes/archive/2026-06-03-fix-flaky-cdc-mixed-mode-test/design.md
@@ -1,0 +1,94 @@
+## Context
+
+`TestCDCBackends/.../Mixed_Mode` (`pkg/cache/cdc_test.go:284`) intermittently fails at line
+316 with `error fetching the nar from the store: not found`. The test stores a whole-file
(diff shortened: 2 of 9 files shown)


kalbasit:

- #1324 ◀

PR #1325, opened 2026-06-04
fix(cache): don't route serve to chunks without a chunk store

kalbasit:
## Summary

Fixes the flaky `TestCacheBackends/.../RunLRUCleanupInconsistentNarInfoState`
failure in CI ([run 26929993682](https://github.com/kalbasit/ncps/actions/runs/26929993682/job/79447676107))
— `chunk store not initialized, cannot serve NAR from chunks`.

The root cause is a TOCTOU on `hasInStore` in `GetNar`. It reads
`HasNarInStore` once, then re-evaluates servability via `isServable`, which
performs its **own** fresh `HasNarInStore`. When the whole-file NAR lands in the
store between those two checks, the first flag is stale (`false`) while the NAR
is in fact present. `serveNarFromStorageViaPipe` then computed
`serveFromChunks := !hasInStore = true` and routed an uncompressed request to
`getNarFromChunks`, which hard-fails when no chunk store is configured.

This is the inverse of #1324 (present-at-check, deleted-before-use); here it is
absent-at-check, present-before-use.

## Fix

- Gate the chunk route on chunk-store availability:
  `serveFromChunks := !hasInStore && c.isChunkStoreAvailable()`.
- When no chunk store exists, an uncompressed request resolves against the
  whole-file store: a present whole file is served; a genuine absence returns a
  clean `storage.ErrNotFound`, preserving cache-miss recovery and the
  upload-only contract.
- When a chunk store **is** present the gate reduces to the original
  `!hasInStore`, so standard CDC and drain-mode serving are byte-identical
  (the #1324 store→chunks fallback is unchanged).

## Test plan

- [x] New deterministic regression test reproduces the exact CI error
  (RED → GREEN): `pkg/cache/cdc_serve_stale_hasinstore_internal_test.go`
- [x] Formerly-flaky `RunLRUCleanupInconsistentNarInfoState` passes 5× under `-race`
- [x] `#1324` migration-race tests still pass (no regression)
- [x] `task fmt`, `task lint` (0 issues), `task test`, and full `task test:auto`
  (real Postgres/MySQL/Redis/Garage) pass

## Spec

Adds the `cdc-chunking` requirement "Serving a whole-file NAR MUST be resilient
to a stale time-of-check store-presence observation" (4 scenarios), and archives
the OpenSpec change under
`openspec/changes/archive/2026-06-03-fix-flaky-pg-test/`.

### DIFF
--- a/openspec/changes/archive/2026-06-03-fix-flaky-pg-test/.openspec.yaml
+++ b/openspec/changes/archive/2026-06-03-fix-flaky-pg-test/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-04
--- a/openspec/changes/archive/2026-06-03-fix-flaky-pg-test/design.md
+++ b/openspec/changes/archive/2026-06-03-fix-flaky-pg-test/design.md
@@ -1,0 +1,118 @@
+## Context
+
+`GetNar` (`pkg/cache/cache.go`, ~line 1085) reads store presence into a local
+`hasNarInStore` once, then determines servability through `isServable`, which
+performs its **own** fresh `HasNarInStore` call. These are two distinct
+time-of-check observations of the same store:
+
+```go
+hasNarInStore := c.HasNarInStore(ctx, narURL)      // (1) may observe absent
+...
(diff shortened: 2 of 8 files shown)


kalbasit:

- #1325 ◀

PR #1326, opened 2026-06-04
fix(cache): non-destructive narinfo purge on root path

kalbasit:
## Summary

Fixes the persistent `nix copy --to .../upload` abort: `cannot add 'X' because the reference 'Y' does not exist`. Root cause confirmed live (2 non-sticky replicas, whole-file local backend on shared NFS, shared Postgres): the **substituter/root** narinfo read path still called `purgeNarInfo` when a backing NAR looked absent. Because the DB is shared across replicas, that delete is globally visible instantly, flipping a verified reference `200 -> 404` mid-copy. #1321 made only the `/upload` path non-destructive.

- **Non-destructive root reads** — `getNarInfoFromDatabase` and `getNarInfoFromStore` now return the cache-miss sentinel on a missing NAR instead of purging; the existing upstream re-fetch heals the record in place. Client-visible HTTP is unchanged (200 upstream-hit, 404 miss, never 500); the destructive cross-replica delete window is gone.
- **Refcount-aware deletion** — `narinfo`↔`nar_file` is M:N (many narinfos → one NAR). `purgeNarInfo` now deletes a `nar_file` record and its NAR bytes only when no other narinfo still links it (orphan-only, mirroring `RunLRU`). Previously a false-positive purge could delete a present, still-shared NAR and break sibling narinfos.
- Removed the now-unused, refcount-unsafe `deleteNarFromStore` helper.
- Modifies the `narinfo-purge-serving` spec (non-destructive substituter re-fetch + refcount-safe deletion).

Not addressed here (tracked separately): the topology itself (whole-file local backend on shared NFS + non-sticky replicas). This change makes phantoms non-destructive and self-healing rather than copy-aborting; the durable fix is an object-store NAR backend.

## Test plan

- [x] `task fmt`, `task lint` (0 issues), `task test` (full suite, `pkg/cache` under `-race`) all pass
- [x] New tests: substituter DB/store reads non-destructive; nar_file record preserved; refcount-safe purge (shared survives / orphan reclaimed); upstream heal-in-place
- [x] Updated the prior `NonUploadOnlyStillPurges` test to assert the new non-destructive behavior

### DIFF
--- a/openspec/changes/archive/2026-06-04-non-destructive-narinfo-purge/.openspec.yaml
+++ b/openspec/changes/archive/2026-06-04-non-destructive-narinfo-purge/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-04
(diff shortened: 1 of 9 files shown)


kalbasit:

- #1326 ◀
    - #1330
        - #1332
            - #1333
                - #1334
                    - #1335

PR #1330, opened 2026-06-05
fix(cache): durable-bytes marker for /upload reference presence

kalbasit:
## Summary

Stacked on #1326. Fixes `nix copy --to .../upload` aborting with
`cannot add X because the reference Y does not exist` against a
production cache carrying CDC-era residue.

Two parts:

1. **`bytes_stored_at` marker** — a dedicated `nar_file` column (distinct
   from fsck's `verified_at`) that `PutNar` sets once a NAR's bytes are
   durably written. On the `/upload` path the narinfo presence check
   trusts this shared-DB marker over the local filesystem stat, so a
   reference check on one replica no longer 404s a NAR a peer replica
   just uploaded but whose write this replica's NFS mount hasn't observed
   yet. A byte-less placeholder row (created by a narinfo PUT) is **not**
   trusted, so phantoms stay excluded.

2. **Compression-agnostic presence lookup** — the marker is matched by
   NAR hash (+query) regardless of compression. Prod has narinfos that
   advertise `url=nar/<hash>.nar` (Compression:none) while the NAR is
   stored only under xz; a compression-keyed lookup missed the xz row and
   wrongly 404'd. The hash identifies the NAR and the marker proves its
   bytes exist, so presence no longer depends on which compression the
   narinfo URL advertises.

Why the `/upload` path only: the substituter (root) path must keep
self-healing a genuinely missing NAR from upstream; trusting the marker
there could mask a real eviction.

## Test plan

- [x] New internal tests: compression-agnostic presence, marker-trust on
      upload path, placeholder-not-trusted phantom guard
- [ ] Deploy to prod and confirm `nix copy --to .../upload` of the
      previously-failing closures succeeds


kalbasit:

- #1326
    - #1330 ◀
        - #1332
            - #1333
                - #1334
                    - #1335

PR #1332, opened 2026-06-05
fix(cache): close the CDC url=none/xz-NAR desync (prevention + repair)

kalbasit:
## Summary

Stacked on #1330. Root-cause fix + data repair for the production
`nix copy --to .../upload` aborts (`cannot add X because the reference Y
does not exist`).

**Root cause** (confirmed against the live CDC snapshot DB): `pullNarInfo`
rewrote the persisted narinfo URL to `nar/<hash>.nar` (Compression none)
+ nulled FileHash **synchronously, before** background eager-CDC chunking
ran. The nar_file is first written whole-file under its real compression
(xz, `total_chunks=0`) and only flipped to none/chunked later by
`SetTotalChunks`. A crash/restart in that window strands the narinfo
advertising a `none` URL while the NAR is stored xz-only at
`/nar/<hash>.nar.xz` → the none URL 404s → reference check aborts. The
broken narinfos' URL hash is the **xz file hash, not the NAR hash**, and
`file_hash` is NULL — the fingerprint of this exact path.

### Prevention (commit 1)
Persist the narinfo **truthfully**. Only normalize to `none` for genuinely
`Compression:none` upstreams (stored zstd, served none transparently — the
Harmonia path, unchanged). Do **not** predict `none` for CDC. Serve-time
`maybeCDCNormalizeNarInfoURL` already presents `url=none` once the NAR is
genuinely chunked (`HasNarInChunks`), so the steady-state serve contract
is unchanged — only the pre-chunk window now advertises the truthful
compression instead of a not-yet-true `none`.

Two CDC contract tests that asserted the removed store-time "normalize to
none" behavior are rewritten to assert the **narinfo↔storage consistency
invariant** instead; the phantom-recovery test now addresses the NAR via
the URL the narinfo advertises.

### Repair (commit 2)
Forward-only per-dialect data migration restoring the **127** production
narinfos already in this state: rewrite `url`→`.nar.xz`, `compression`→xz,
`file_hash`→`'sha256:'||nar_file.hash`, `file_size`→nar_file.file_size,
all reconstructed from the joined nar_file. Narinfos that also have a
servable none/chunked backing are excluded (`NOT EXISTS`). Idempotent;
no-op on clean DBs. Validated live against the CDC snapshot (UPDATE 3,
second run 0).

## Test plan

- [x] New internal test: eager-CDC xz pull persists the truthful xz URL
- [x] CDC consistency tests + phantom-recovery green (`-count=3`, non-flaky)
- [x] Repair UPDATE validated live on the CDC snapshot DB (3 rows, idempotent)
- [ ] Deploy + confirm the 127 previously-broken paths serve once the
      migration applies


kalbasit:

- #1326
    - #1330
        - #1332 ◀
            - #1333
                - #1334
                    - #1335

PR #1333, opened 2026-06-05
fix(cache,fsck): reconcile narinfo<->nar_file links (drain-mode unblock + prevention)

kalbasit:
## Summary

Stacked on #1332. Roots out the cause of the cache being stuck in CDC
**drain mode** (2,499 chunked NARs the de-chunk pass silently skipped),
and removes a related **fsck data-loss hazard**.

**Root cause:** the `narinfo_nar_files` join link is created in the
narinfo-write path (`storeInDatabase`), decoupled from the async CDC
chunking that finalizes the chunked `nar_file`. A rare race (~0.8% of CDC
NARs in a prod snapshot) leaves the chunked `nar_file` **unlinked**. That
link is the *only* way `linkedNarinfoNarHash` resolves the verification
hash, so de-chunk skipped every unlinked NAR (`ErrNoNarHashToVerify`) →
drain never exits. Worse, fsck classified those valid, reachable
narinfos as orphans and **deleted** them.

Three fixes (A+C in `cache.go`, B in `fsck.go`):

- **A — de-chunk robustness:** `linkedNarinfoNarHash` falls back to
  resolving the narinfo by the NAR's `Compression:none` URL when no join
  link exists, so an unlinked chunked NAR is still content-verified and
  drained.
- **C — creation-race prevention:** `checkAndFixNarInfosForNar` (runs
  right after chunking completes) reconciles the link via upsert — a
  no-op when it already exists, closing the race at the source.
- **B — fsck safety:** repair the missing link (recreate it when the
  nar_file the URL references is present) instead of deleting the
  narinfo; only genuine orphans are still reclaimed.

A follow-up data remediation (backfill/repair the existing 2,499, re-run
`migrate-chunks-to-nar`, exit drain) is performed operationally against
prod with this image.

## Test plan

- [x] de-chunk of an unlinked chunked NAR succeeds via URL fallback
- [x] chunking-completion reconciles the missing link
- [x] fsck repairs an unlinked narinfo with a backing nar_file; still
      deletes one with no backing (`-count=3`, non-flaky)
- [ ] prod: re-run de-chunk drains all chunked NARs, then exit drain mode


kalbasit:

- #1326
    - #1330
        - #1332
            - #1333 ◀
                - #1334
                    - #1335

PR #1334, opened 2026-06-05
fix(cache,fsck): self-completing CDC drain (de-chunk never strands NARs)

kalbasit:
## Summary

Stacked on #1333. Makes `migrate-chunks-to-nar` always drive the chunked
count to zero so CDC drain auto-completes without manual SQL, and adds an
fsck residue-repair safety net.

Three production failure classes (kept un-purged as a real-corruption
fixture):
1. ~112 chunked NARs whose only NarHash-bearing narinfo advertises a
   different-compression URL (`.nar.xz`) than the bare `nar/<H>.nar`, so
   the verify-hash resolver misses and **skips** them.
2. De-chunking such a NAR to none/whole without normalizing its narinfo
   URL recreates the `.nar.xz` ≠ none/whole 404.
3. ~4 NARs whose reconstruction hard-fails are counted `failed` and left
   chunked.

Code changes (proposal/design/specs in `openspec/changes/`):
- **(A)** resolve the verify NarHash by NAR hash from any referencing
  narinfo, not just the exact none URL;
- **(B)** normalize the narinfo URL to none on successful de-chunk;
- **(C)** purge-on-unverifiable (no NarHash / hard reconstruction fail)
  instead of skip/fail, so the pass always reaches `chunked = 0`;
- **(D)** an fsck repair pass that relinks/normalizes/purges chunked-NAR
  residue.

This PR currently contains the OpenSpec proposal only; implementation
follows (validated against the real production stragglers and the
CDC-lifecycle e2e tests).

## Test plan

- [ ] TDD fixtures per failure class (`.nar.xz`-URL narinfo, no-NarHash, corrupt chunks)
- [ ] full pass over a mixed fixture leaves `chunked = 0`
- [ ] deploy + run against the real 116 stragglers; drain auto-completes


kalbasit:

- #1326
    - #1330
        - #1332
            - #1333
                - #1334 ◀
                    - #1335

PR #1335, opened 2026-06-05
feat(fsck): reclaim CDC chunked-NAR residue (grace-period mark-then-purge)

kalbasit:
## Summary

A routine-fsck safety net for CDC chunked-NAR residue, complementing the
self-completing de-chunk migration (#1334).

Two tiers, per chunked `nar_file`:
- **Recoverable** (a narinfo with a resolvable NarHash, inconsistent URL): relink + normalize the URL to none **immediately** (touches no chunks; safe in any CDC state).
- **Un-de-chunkable** (no resolvable NarHash): **mark** with a persistent `nar_file.dechunk_residue_flagged_at` timestamp; **purge only on a later fsck run** once the flag has aged past a configurable ~24h grace window (`--dechunk-residue-grace`) and the row is still suspect. Becomes-recoverable/de-chunked → flag cleared, never purged. Rows mid-chunking are skipped.

The grace window (two daily runs) prevents purging transient states and protects legitimately chunked NARs. Non-repair/dry-run runs only report already-flagged, grace-elapsed residue, so they raise no false positives.

Includes the OpenSpec proposal **and** the implementation: nullable `dechunk_residue_flagged_at` column (forward-only additive migration, all three dialects), the two-tier reclaimer in `pkg/ncps/fsck.go` reusing `PurgeChunkedNar` and the shared de-chunk resolver, a white-box policy test, the five spec scenarios as CLI tests, and a CDC-lifecycle e2e exercising the reclaimer alongside the drain.


kalbasit:

- #1326
    - #1330
        - #1332
            - #1333
                - #1334
                    - #1335 ◀

PR #1339, opened 2026-06-07
fix(ent): stop sqlite phantom narinfos/nar_files rebuild

### LINKED ISSUE #1328: sqlite migration generator spuriously rebuilds the unrelated `narinfos` table on every new migration
## Summary

`task migrations:gen` produces, **for the sqlite dialect only**, a full table-rebuild of the unrelated `narinfos` table on *every* new migration — even when the schema change only touches a different table (e.g. adding a column to `nar_files`).

## Symptom

Generating a migration that only adds `nar_files.bytes_stored_at` yields sqlite output that, in addition to rebuilding `nar_files`, also does a full swap of `narinfos`:

```sql
PRAGMA foreign_keys = off;
-- create "new_narinfos" table  (identical columns to the current narinfos)
CREATE TABLE `new_narinfos` ( ... CONSTRAINT `narinfos_file_size_nonneg` CHECK (file_size >= 0), CONSTRAINT `narinfos_nar_size_nonneg` CHECK (nar_size >= 0));
INSERT INTO `new_narinfos` (...) SELECT ... FROM `narinfos`;
DROP TABLE `narinfos`;
ALTER TABLE `new_narinfos` RENAME TO `narinfos`;
-- recreate indexes ...
-- then the same swap for nar_files ...
```

Postgres and MySQL output for the same change is a clean `ALTER TABLE nar_files ADD COLUMN ...` — only sqlite is affected.

## What I ruled out

The `narinfos` CHECK constraints atlas emits (`narinfos_file_size_nonneg`, `narinfos_nar_size_nonneg`) **are already present** in the committed sqlite baseline (`migrations/sqlite/20260521021523_ent_baseline.sql` and `20260525175108_add_build_trace_entries.sql`). So this is not a missing-constraint case — atlas is detecting a subtler normalization difference between the schema produced by replaying the sqlite migrations and the Ent-desired schema for `narinfos` (e.g. CHECK-expression text/quoting/whitespace normalization).

## Impact

- Every new sqlite migration must be **hand-trimmed** down to the intended change to avoid committing a spurious, heavy `narinfos` table-swap. This was done for the `verified_at` migration previously, and again for `bytes_stored_at` in #1326.
- Risk that a future migration is committed with the unintended `narinfos` rebuild (destructive table swap) for sqlite deployments.
- Undermines trust in the generator output for sqlite; encourages manual editing (which is otherwise discouraged).

## Repro

1. Add any field to a non-`narinfos` table in `ent/schema/`.
2. `task ent:generate && task migrations:gen NAME=<x>`.
3. Observe `migrations/sqlite/*_<x>.sql` rebuilds `narinfos` too.

## Suggested next step

Inspect the exact diff atlas computes between the replayed sqlite baseline and the Ent desired schema (e.g. `atlas schema diff` / `atlas migrate diff` with verbose output) to pinpoint the `narinfos` normalization mismatch, then add a one-time corrective sqlite migration so future diffs are clean.

_Detected while implementing the multi-replica NAR-presence fix (#1326)._

kalbasit:
## Summary

Fixes #1328.

`task migrations:gen` rebuilt the unrelated `narinfos` and `nar_files`
tables on **every** new sqlite migration, forcing each migration to be
hand-trimmed and risking a committed destructive table swap.

**Root cause** (traced via Ent's `WithDiffHook`): `last_accessed_at`
declared its DB default with `entsql.Annotation{DefaultExpr:
"CURRENT_TIMESTAMP"}`, which Ent emits as a parenthesized Atlas
`RawExpr`. Atlas's sqlite inspector strips the parens on readback, so
desired (`(CURRENT_TIMESTAMP)`) and inspected (`CURRENT_TIMESTAMP`)
never compare equal — a perpetual `ModifyColumn` (ChangeDefault) that
sqlite expresses as a full table rebuild. `created_at` was immune
because the `Timestamps` mixin already uses the string form
`entsql.Default("CURRENT_TIMESTAMP")`. The issue's suggested corrective
migration would **not** have worked: the stored DDL is already
atlas-canonical; the mismatch is in Ent's desired-schema representation.

## What changed

- `ent/schema/{narinfo,nar_file}.go`: `last_accessed_at` now uses
  `entsql.Default("CURRENT_TIMESTAMP")`; regenerated the Ent client.
  Emitted DDL (`DEFAULT (CURRENT_TIMESTAMP)`) is unchanged, so **no
  migration is needed** and existing databases are untouched.
- `cmd/ent-lint`: new invariant **A6** rejecting the `DefaultExpr`
  form, to prevent regression (wired into `nix flake check`).
- `cmd/generate-migrations`: `TestSQLiteNoPhantomDiff` pinning the
  empty-diff contract (RED before the fix, GREEN after).
- `.claude/rules/ent-migrations.md`: documented the convention (§6).
- Synced the `database-migrations` and `ent-schema-lint` specs.

## Test plan

- [x] `task fmt`, `task lint` (0 issues), `task test` (29 pkgs, 0 fail)
- [x] `task ent:check` exits 0 (A6 PASS on all schemas; ent drift clean)
- [x] `task migrations:gen` produces no spurious migration across
  **sqlite, postgres, and mysql** (verified against live dev DBs)
- [x] `git status` shows zero changed files under `migrations/`

### DIFF
--- a/.claude/rules/ent-migrations.md
+++ b/.claude/rules/ent-migrations.md
@@ -87,6 +87,29 @@ This is enforced at the source level (AST check on the schema file) by
 `task uar:ent:lint`. The `Sensitive()` annotation is a schema-only metadata
 change; it does NOT alter the SQL column definition and requires no migration.
 
+### 6. `CURRENT_TIMESTAMP` DB defaults MUST use `entsql.Default(...)`, never `DefaultExpr`
+
+A DB-level `CURRENT_TIMESTAMP` default MUST be declared via
+`entsql.Default("CURRENT_TIMESTAMP")` (which Ent emits as a string default that
+Atlas's SQLite inspector round-trips exactly). It MUST NOT be declared via
+`entsql.Annotation{DefaultExpr: "CURRENT_TIMESTAMP"}`: Ent emits that form as a
+parenthesized Atlas `RawExpr`, but Atlas's SQLite inspector strips the parens on
+readback, so the desired (`(CURRENT_TIMESTAMP)`) and inspected
+(`CURRENT_TIMESTAMP`) values never compare equal. The result is a perpetual
+phantom `ModifyColumn` (ChangeDefault) that rebuilds the whole table on **every**
+generated SQLite migration (GitHub issue #1328).
+
+```go
+field.Time("last_accessed_at").
+    Optional().
+    Nillable().
+    Default(time.Now).
+    Annotations(entsql.Default("CURRENT_TIMESTAMP")), // NOT entsql.Annotation{DefaultExpr: ...}
+```
+
+This is enforced at the source level (AST check, invariant A6) by `cmd/ent-lint`.
+It is a schema-representation-only change; the emitted SQL DDL (`DEFAULT (CURRENT_TIMESTAMP)`) is unchanged, so it requires no migration.
+
 ## Snake_case enum types
 
 When a `field.Enum(...)` field generates a Postgres ENUM type, give it an
--- a/cmd/ent-lint/main.go
+++ b/cmd/ent-lint/main.go
@@ -11,6 +11,11 @@
 //   - A4: every edge.To declaration in any schema must have a reciprocal
 //     edge.From(...).Ref(...) on the target schema (otherwise Ent fabricates
 //     a phantom FK column on the target).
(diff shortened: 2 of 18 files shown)


kalbasit:

- #1339 ◀

PR #1341, opened 2026-06-07
fix(cache): delete the stored variant on NAR eviction

kalbasit:
## Summary

`Cache.DeleteNar` removed only the exact URL variant via `narStore.DeleteNar`, but a `Compression:none` NAR's physical object is stored under the `.nar.zst` variant. Deleting `nar/<H>.nar` therefore left the real `.nar.zst` blob orphaned on disk while still clearing `bytes_stored_at` — leaking storage and making the `/upload` presence check lie (report absent while bytes remain). It also spuriously errored, since the bare none object never exists.

This deletes the actually-stored variant (the `.nar.zst` variant for `none`, mirroring `statNarInStore`) and tracks whether any variant was present, returning `ErrNotFound` only when none was — preserving the established "deleting an absent NAR errors" contract while no longer orphaning the blob.

Surfaced by a CodeRabbit "outside diff range" finding on #1331; scoped out of that PR as pre-existing, now fixed on its own.

Specs: adds a requirement to `upload-reference-presence` (deletion removes the stored variant before clearing the marker; not-found only when no variant was present).

## Test plan

- [x] `task fmt`, `task lint`, `task test` pass
- [x] New: `TestDeleteNar_RemovesNoneNarZstdVariant` (none NAR removes the `.nar.zst` object)
- [x] New: `TestDeleteNar_AbsentNoneNarReturnsNotFound` (contract: absent → `ErrNotFound`, marker kept)
- [x] Existing xz `DeleteNar` "absent → ErrNotFound" backend test still passes

### DIFF
--- a/openspec/changes/archive/2026-06-06-fix-deletenar-orphaned-bytes/.openspec.yaml
+++ b/openspec/changes/archive/2026-06-06-fix-deletenar-orphaned-bytes/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-07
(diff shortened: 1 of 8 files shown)


kalbasit:

- #1341 ◀
    - #1342

PR #1343, opened 2026-06-07
fix(metrics): prime counters for startup exposure

### LINKED ISSUE #1337: Missing prometheus metrics
According to the docs there should be more metrics exposed under `/metrics` when prometheus is enabled.

I only can see:
 - `target_info`
 - `requests_inflight`
 - `http_client_request_body_size_bytes_{bucket,count,sum}`
 - `http_client_request_duration_seconds_{bucket,count,sum}`
 - `request_duration_millis_milliseconds_{bucket,count,sum}`
 - `response_size_bytes_{bucket,count,sum}`

I seem to be missing at least:
 - `ncps_nar_served_total`
 - `ncps_narinfo_served_total`

Running on `NixOS`, version `0.9.4`.

kalbasit:
## Summary

Fixes #1337. Documented counters (`ncps_nar_served_total`,
`ncps_narinfo_served_total`, the LRU eviction counters, the lock
counters, and the migration counter) were absent from `GET /metrics`
on a freshly started or idle instance.

**Root cause:** the OpenTelemetry → Prometheus exporter only emits a
counter time series after the counter records its first measurement.
These counters only increment on real events, so until a NAR/narinfo
was served (or an eviction/migration fired) the series simply did not
exist — contradicting the docs that promise them and breaking
dashboards/alerts that reference them from the start.

**Fix:** add an exported `PrimeMetrics(ctx)` to `pkg/cache`,
`pkg/lock`, and `pkg/ncps` that records a zero-valued `Add(ctx, 0)` on
each `Int64Counter`, and call all three from `serve.go` once the global
meter provider is installed (when OTel or Prometheus is enabled).
Adding zero forces each series to export at value `0` without affecting
counts: the first real event still advances the counter from 0 to 1.
Observable gauges and histograms are unaffected (gauges already export
via callbacks).

## Test plan

- [x] `task fmt`, `task lint` (0 issues), `task test` (race detector) all pass
- [x] New `pkg/ncps/metrics_prime_test.go` drives the real Prometheus
      gatherer and asserts all 13 counters appear at `0` before any
      traffic, plus a 0→1 increment check proving priming does not inflate
- [x] `pkg/cache` and `pkg/lock` no-op-without-provider tests confirm
      priming is safe when metrics are disabled

### DIFF
--- a/openspec/changes/archive/2026-06-06-fix-prometheus-counter-priming/.openspec.yaml
+++ b/openspec/changes/archive/2026-06-06-fix-prometheus-counter-priming/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-07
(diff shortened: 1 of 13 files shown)


kalbasit:

- #1343 ◀

PR #1346, opened 2026-06-07
fix(cache): serve whole-file NAR with its true compression under CDC

kalbasit:
## Summary

`getNarFromStore` streams the raw bytes of the whole file it opened (e.g. `xz`), but then overwrote `narURL.Compression` with the compression of the record returned by `getNarFileFromDB`, which is **CDC-first** and returns the coexisting chunked `compression=none` row during the lazy-chunking transition window. The response then advertised `Compression: none` while the body was raw xz, with `Content-Length` set to the xz size (< `NarSize`), so `nix` failed with:

- `error: NAR for '…' fetched from … is incomplete`
- `error: input compression not recognized`

This deterministically broke the **enable-CDC-on-an-existing-whole-file-cache** upgrade path. It is the NAR-serve counterpart to the narinfo-side normalization already done in `maybeCDCNormalizeNarInfoURL` (#1332).

## Fix

Query the `nar_file` row for the representation actually being served (`hash + served compression + query`) instead of the CDC-first lookup, and stop overwriting `narURL.Compression`. The served compression now matches the streamed bytes; the `last_accessed_at` touch targets the served row.

Adds a `pkg/cache` regression test (`TestCDCWholeFileServeReportsServedCompression`) and the OpenSpec change `fix-cdc-incomplete-nar-serve` (adds a `cdc-chunking` requirement: GetNar must advertise the compression of the bytes it serves).

## Test plan

- [x] New regression test fails before the fix (returns `none`, short length) and passes after
- [x] `task fmt`, `task lint`, full `task test` green
- [x] Caught + then validated end-to-end by the new CDC lifecycle e2e driver (top of stack)


kalbasit:

- #1326
    - #1330
        - #1332
            - #1333
                - #1334
                    - #1346 ◀
                        - #1347
                            - #1348

PR #1347, opened 2026-06-07
fix(docker): ship /etc/passwd and /etc/group as regular files

kalbasit:
## Summary

The image added `/etc/passwd` and `/etc/group` via a `buildEnv`, which makes them **absolute symlinks into the nix store** (`/etc/passwd -> /nix/store/…-passwd/etc/passwd`). Recent containerd securejoins `etc/passwd` during container creation and rejects the absolute symlink as escaping the rootfs, so **every** ncps pod fails to start:

```
failed to create containerd container: … openat etc/passwd: path escapes from parent
```

This is a latent **production** failure on recent k8s/containerd (reproduced on `kindest/node:v1.35.0`), and it blocked the entire `k8s-tests` Kind suite.

## Fix

Materialize `/etc/passwd` and `/etc/group` as **real files** in `fakeRootCommands` (re-materializing the read-only store-symlinked `/etc` as a writable real directory first), with identical content (`root` + `ncps` uid/gid 1000). Adds a `checks.docker-image-etc-files` regression guard asserting both are regular files (not symlinks).

OpenSpec change: `fix-docker-etc-passwd` (new capability `container-image-etc-files`).

## Test plan

- [x] `checks.docker-image-etc-files` builds green (both are regular files)
- [x] Before/after on Kind: old (symlinked) image → `CreateContainerError` with the exact error; fixed image → pod `Completed`
- [x] `disallowedRequisites` (no shell/coreutils) still holds; `task fmt`/`lint` clean


kalbasit:

- #1326
    - #1330
        - #1332
            - #1333
                - #1334
                    - #1346
                        - #1347 ◀
                            - #1348

PR #1349, opened 2026-06-07
test(cache): de-flake TestGetNarInfoOpaqueUpstreamURL size assert

kalbasit:
## Summary

`TestGetNarInfoOpaqueUpstreamURL` flaked in CI with `cache_test.go:529 expected: 50308, actual: -1`. The first `GetNar` asserted the exact NAR size, but `GetNar` legitimately returns `-1` (size unknown) when it serves bytes via the live-streaming path while the upstream download is still in flight (`cache.go:1449`). Which path it takes is timing-dependent, so under CPU contention (parallel integration cohorts) the assertion failed even though the streamed bytes were always correct.

- **Fix (`pkg/cache/cache_test.go`):** accept either the concrete file size or `-1`, keeping the authoritative byte-equality check — mirroring the post-eviction re-fetch just below that already handled this.
- **Spec (`nar-concurrent-streaming`):** document the `GetNar` `-1` streaming size contract so it's explicit and testable.
- **Spec (`api-surface`):** correct the stale `Cache.GetNar` / `NarStore` / `NarInfoStore` signatures to match the real code (the block also omitted `StatNar`/`WalkNars` and listed a phantom `GetNarSize`). Docs only — no behavior change.

OpenSpec change archived as `2026-06-07-fix-flaky-opaque-upstream-nar-size`; canonical specs synced.

## Test plan

- [x] Reproduced the flake deterministically: `go test ./pkg/cache -run '^TestGetNarInfoOpaqueUpstreamURL$' -count=60 -race` under full CPU saturation → `expected 50308, actual -1`.
- [x] After the fix, 80× `-race` runs under full CPU load → `ok`.
- [x] `task fmt` (0 changed), `task lint` (0 issues), `task test` (pass).
- [x] `openspec validate --specs` → 33 passed, 0 failed.


kalbasit:

- #1349 ◀

PR #1350, opened 2026-06-07
fix(cache): delete dangling narinfos when GC'ing genuinely-absent NAR

kalbasit:
## Summary

The recovery sweep's `gcOrSkipBackingLessNarFile` deletes a backing-less `nar_file` when **every** linked narinfo is genuinely absent from **every** healthy upstream. The `narinfo_nar_files` on-delete cascade removed only the link rows and left the parent narinfos behind — **dangling**, linked to no `nar_file`.

Such a narinfo was already served as HTTP 200 and cached by Nix clients, so a later `nix build` fetches the NAR directly (skipping any narinfo re-fetch) and gets HTTP 404:

```
error: file 'nar/<hash>.nar' does not exist in binary cache
```

Prod (`v0.10.0-rc10`) accumulated **2,744** such dangling narinfos. The read-path guard (`narinfo-purge-serving`) cannot help — it only fires on a narinfo *re-request*, which a Nix client-cache hit bypasses.

## Fix

In the genuinely-absent branch, delete the `nar_file` **and** its linked narinfos atomically in one Ent transaction (`nar_file` first so its cascade clears the links, then the now-unlinked narinfos).

The conservative gate is unchanged: any `Present`/`Unknown` upstream, no healthy upstreams, the zero-linked-narinfo branch, and fresh placeholders all behave exactly as before. No schema or migration change.

Implements the new requirement *"The backing-less GC MUST NOT leave a dangling narinfo"* in spec `nar-cache-miss-recovery` (change archived as `2026-06-07-prevent-dangling-narinfos`).

## Test plan

- [x] TDD red→green: new genuinely-absent (single + multi-narinfo) tests assert narinfos are deleted; strengthened present/unknown guards assert narinfos are kept; added zero-linked regression
- [x] `task fmt` (0 changed), `task lint` (0 issues), `task test` (full suite passes, race detector)
- [x] No `ent/` or `migrations/` drift

### DIFF
--- a/openspec/changes/archive/2026-06-07-prevent-dangling-narinfos/.openspec.yaml
+++ b/openspec/changes/archive/2026-06-07-prevent-dangling-narinfos/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-07
--- a/openspec/changes/archive/2026-06-07-prevent-dangling-narinfos/design.md
+++ b/openspec/changes/archive/2026-06-07-prevent-dangling-narinfos/design.md
@@ -1,0 +1,49 @@
+## Context
+
+The recovery sweep calls `gcOrSkipBackingLessNarFile` (`pkg/cache/cache.go`) for a `nar_file` whose whole-file bytes are gone and which migration cannot help. Current logic:
+
+1. Query narinfos linked via `narinfo_nar_files` to this `nar_file`.
+2. If **zero** linked narinfos → delete the `nar_file` (clean GC). Correct.
+3. If **≥1** linked narinfo and **every** one is `narInfoGenuinelyAbsentUpstream` (confirmed gone from every healthy upstream; conservative — any Present/Unknown/no-upstreams aborts) → delete the `nar_file`.
+4. Otherwise → skip; leave for on-demand `GetNar` recovery.
+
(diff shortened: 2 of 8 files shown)


kalbasit:

- #1350 ◀

kalbasit:
Addressed both review points in 6d614125:

- **@coderabbitai (outside-diff race):** confirmed real. The linked-narinfo set was snapshotted before the network-bound upstream-absence probes, so a concurrent `storeInDatabase`/repair could link a new narinfo to the same `nar_file` before the delete — the `nar_file` delete then cascaded the new link away and the snapshot delete missed the narinfo, re-creating the dangling row (and the sweep could no longer find it). Reworked to delete **only the verified narinfos and their links**, then GC the `nar_file` **only if no links remain**. A concurrently-linked narinfo keeps its link, keeps the row referenced, and is never orphaned; the next sweep re-evaluates with a fresh, fully-verified snapshot. Isolation-safe without row locks (the `sql/lock` Ent feature isn't enabled here). Extracted into `gcDeleteAbsentNarInfosAndMaybeNarFile` so the guard is deterministically unit-tested.
- **@gemini-code-assist (N+1):** replaced the per-row delete loop with a single bulk `Delete().Where(IDIn(...))`.

No schema/migration change. `fmt`/`lint`/`test` (race) all green.

Issue #660, opened 2026-01-27
ncps does not handle NAR streaming in a multi-instance HA setup


kalbasit:
## Problem statement

In a multi-instance HA setup with Redis locks:
- Instance A starts downloading a large NAR file from upstream
- Instance B receives a client request for the same NAR file
- Instance B acquires the Redis download lock and sees the download is in progress
- Instance B cannot stream from Instance A's in-memory download state
- Instance B must wait for Instance A to complete and store the file
- For large files, this wait exceeds client timeouts (HTTP 200 with curl timeout errors)

Extracted from closed #618:

> We're seeing this as well, and not even with heavy loads, sometimes just with a single build. It does not appear to be resolved in `v0.7.3`.
> 
> I've dropped our `ncps` deployment back to 1 instance to see if that helps. FYI, we're using:
> 
> * PostgreSQL
> * NVMe-backed S3 storage via Ceph RGW
> * Redis locking via DragonflyDB 

 _Originally posted by @dhess in [#618](https://github.com/kalbasit/ncps/issues/618#issuecomment-3802292166)_

kalbasit:
From my conversation with Claude, I can either fix this short-term and target 0.7+ or fix it long-term and just get #322 done for 0.9 instead of what I had planned for 0.10. I'm inclined to solve this by implement CDC earlier than planned. LMK @dhess if this is breaking your side of things, it will probably take a week give or take to implement, test and release especially considering that I have yet to release 0.8.

---

# Fix Issue #660 - Multi-Instance HA Timeout During Downloads

## Problem Statement

In a multi-instance HA setup with Redis locks:
- **Instance A** starts downloading a large NAR file from upstream
- **Instance B** receives a client request for the same NAR file  
- **Instance B** sees the download locked in Redis
- **Instance B** cannot access instance A's in-memory [downloadState](file:///Users/wnasreddine/code/repositories/github.com/kalbasit/ncps/pkg/cache/cache.go#305-333)
- **Instance B** must wait for instance A to complete the download
- For large files, this exceeds client timeouts → "HTTP 200 (curl error: Timeout was reached)"

## Root Cause

The [downloadState](file:///Users/wnasreddine/code/repositories/github.com/kalbasit/ncps/pkg/cache/cache.go#305-333) used for streaming is stored in instance-local memory (`c.upstreamJobs map[string]*downloadState`). When instance B tries to serve the same file, it cannot access instance A's download progress.

## Evaluation of Solutions

[...]

PR #1352, opened 2026-06-07
feat(storage): in-flight staging part-objects

kalbasit:
Add staging part-object storage to the NarStore interface and both
backends (local, S3), the transport layer for in-flight NAR staging
(GitHub #660, #1289). Parts are immutable, fixed-size, sequentially
indexed objects under a per-hash staging namespace that any replica can
read while another is still downloading.

Why: S3 cannot serve an in-progress object, so to stream an in-flight
NAR cross-pod the holder must commit completed immutable pieces during
the download. These methods are the storage substrate; the holder/reader
wiring lands in a later commit.

How:
- NarStore gains PutStagingPart / GetStagingPart / DeleteStagingParts.
- local: parts live at <root>/staging/<hash>/<index>.part, written via
  atomic temp+rename so a part is only readable once fully committed.
- s3: keys <prefix>/store/staging/<hash>/<index>.part; delete lists and
  removes the per-hash prefix. A missing part returns storage.ErrNotFound.

### DIFF
--- a/pkg/storage/local/local.go
+++ b/pkg/storage/local/local.go
@@ -485,6 +485,78 @@ func (s *Store) PutNar(ctx context.Context, narURL nar.URL, body io.Reader, _ in
 	return written, os.Chmod(narPath, fileMode)
 }
 
+// PutStagingPart writes one immutable in-flight staging part-object.
+func (s *Store) PutStagingPart(
+	ctx context.Context,
+	hash string,
+	index int64,
+	body io.Reader,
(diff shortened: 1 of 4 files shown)


kalbasit:

- #1351
    - #1352 ◀
        - #1353
            - #1354
                - #1355
                    - #1356
                        - #1357
                            - #1358
                                - #1359
                                    - #1360
                                        - #1361
                                            - #1362

PR #1355, opened 2026-06-07
feat(cache): prefer staging over progressive chunks

kalbasit:
Read-path precedence for in-flight staging during the eager-CDC chunking
window (#1289, group 7). getNarFromChunks now, when total_chunks == 0 and
staging is enabled, checks for available staging parts and serves the
complete whole-NAR representation from them instead of entering
streamProgressiveChunks. The staged parts are an ordered, complete copy, so
this avoids the fragile progressive reassembly that can truncate.

The precedence is strictly scoped to the chunking window: total_chunks > 0
never reaches the branch, so steady-state chunk serving is unchanged, and
when no staging parts are present the progressive path remains the fallback.
The shared putStagingParts test helper now takes the NarStore interface so the
precedence tests can stage via c.narStore.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -8117,6 +8117,21 @@ func (c *Cache) getNarFromChunks(ctx context.Context, narURL *nar.URL) (int64, i
 		return 0, nil, err
 	}
 
+	// During the chunking window (total_chunks == 0), prefer in-flight staging over
+	// progressive chunk streaming: the staged whole-NAR parts are a complete,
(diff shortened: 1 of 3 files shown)


kalbasit:

- #1351
    - #1352
        - #1353
            - #1354
                - #1355 ◀
                    - #1356
                        - #1357
                            - #1358
                                - #1359
                                    - #1360
                                        - #1361
                                            - #1362

kalbasit:
@coderabbitai review

PR #1372, opened 2026-06-08
feat(dev): add --inflight-staging flag to run.py

kalbasit:
## Summary

- The in-flight NAR staging feature (`serve --cache-inflight-staging-enabled`, added in 541a25c) had no way to be enabled from the dev harness `dev-scripts/run.py`, so it could not be exercised or reproduced locally.
- Adds an `--inflight-staging` toggle that appends `--cache-inflight-staging-enabled` to every spawned `serve` instance. Only the enabled toggle is exposed; `retention` (5m) and `part-size` (8 MiB) keep their Go defaults.
- No locker guard rails are added, mirroring the Go-side guard where the feature self-disables on non-distributed lockers (it only engages with `--locker redis`).
- Effective state is surfaced in the startup banner and persisted to `state.json` (`inflight_staging`) so test drivers can read it.
- Captures the `dev-run-inflight-staging` capability spec (OpenSpec change archived).

## Test plan

- [x] `task fmt` / `task lint` pass
- [x] `run.py --inflight-staging` emits `--cache-inflight-staging-enabled` and records `inflight_staging: true`; without the flag no `--cache-inflight-staging-*` arg appears and state shows `false`

### DIFF
--- a/dev-scripts/run.py
+++ b/dev-scripts/run.py
@@ -490,6 +490,11 @@ def main():
         help="Enable the lazy CDC feature",
     )
     parser.add_argument(
+        "--inflight-staging",
+        action="store_true",
+        help="Enable in-flight NAR staging (only engages with a distributed locker, e.g. --locker redis)",
+    )
+    parser.add_argument(
         "--log-level",
         choices=["debug", "info", "warn", "error", "fatal", "panic"],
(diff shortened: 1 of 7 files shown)


kalbasit:

- #1372 ◀

PR #1373, opened 2026-06-08
test(dev): add inflight-staging contention e2e; drop migration test

kalbasit:
## Summary

- Add `dev-scripts/test-inflight-staging-contention-e2e.py` + `test-inflight-staging-contention-auto.sh` (and a `task test:inflight-staging-contention` target): a multi-replica + Redis-locker contention driver that activates in-flight NAR staging by racing concurrent readers against an in-flight download, asserting every reader gets a complete, byte-identical NAR (vs. `nix-store --dump`) across the download/chunking windows and local/s3 backends. Fails loudly if staging never activates.
- Remove the obsolete `dev-scripts/test-migration-e2e.py` (the dbmate→Ent cutover it validated has shipped; unwired, no importers). The shared `dbmate` create/drop in `run.py`/`migrate-all.py` is retained.

This harness drove the discovery and validation of the in-flight-staging fixes stacked on top of it.

## Test plan

- [x] `task fmt` / `task lint` / `task test` pass
- [x] Driver validated live against Garage/Postgres/Redis (download window green)

PR #1374, opened 2026-06-08
fix(cache): engage in-flight staging under real contention

kalbasit:
## Summary

- The in-flight NAR staging producer opened the holder's download temp file unconditionally; when a cross-pod staging request was observed at download completion, the temp file had already been removed by the post-completion cleanup, so the open failed with ENOENT and the producer logged a WARN error instead of staging — staging never served.
- `produceStagingParts` now treats `fs.ErrNotExist` as a clean no-op (the NAR is already committed to storage; waiters serve from there).
- `stagingActivationPollInterval` lowered 1s → 200ms to match the waiter's cadence, so the holder observes a request while the download is still in flight.

Proven green via the contention e2e in the **download window** across local and s3 (staging activates on the non-holder; byte-identical NARs; no producer WARN).

## Test plan

- [x] `task fmt` / `task lint` / `task test` (race) pass
- [x] New `pkg/cache` unit test (red→green)
- [x] Multi-process e2e: download window green (local + s3)

### DIFF
--- a/dev-scripts/test-inflight-staging-contention-e2e.py
+++ b/dev-scripts/test-inflight-staging-contention-e2e.py
@@ -81,12 +81,14 @@ STAGING_ACTIVATION_LOG = (
 LOCK_CONTENTION_LOG = "failed to acquire download lock"
 PRODUCER_ERROR_LOG = "in-flight staging producer stopped with error"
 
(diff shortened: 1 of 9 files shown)


kalbasit:

- #1373
    - #1374 ◀
        - #1375
            - #1376

PR #1375, opened 2026-06-08
fix(cache): route a contended CDC loser to staging, not a 404

kalbasit:
## Summary

- Under eager CDC, a non-holder replica 404s a compressed NAR request: the download-coordination poll loop early-returned a "served-by-peer" state for an *actively-chunking* NAR because `isServable` conflates "finished" with "actively chunking", routing to chunk serving (which can't produce a compressed variant from decompressed chunks).
- Split a finished-only predicate `hasFinishedNar` (whole-file-in-store OR `total_chunks>0`) out of `isServable`, threaded as `hasFinishedAsset` through `coordinateDownload` + `pollForDownloadOrTakeOver`, and restructured the per-tick decision into four states (dead-holder→takeover[D5] / finished→serve / chunking+staging→staging / chunking+no-staging→progressive).

> **Status:** a correct, regression-free **prerequisite** toward closing the chunking-window cross-pod 404 (#1289). It does NOT fully close that 404 on its own — see the stack head (#1376) for the remaining root cause (a CDC lock-lifecycle design mismatch).

## Test plan

- [x] `task fmt` / `task lint` / `task test` (race) pass — no regression to progressive-chunk/takeover paths
- [x] New `hasFinishedNar` unit test

### DIFF
--- a/openspec/changes/archive/2026-06-08-fix-cdc-window-nonholder-404/.openspec.yaml
+++ b/openspec/changes/archive/2026-06-08-fix-cdc-window-nonholder-404/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-08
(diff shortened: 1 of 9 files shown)


kalbasit:

- #1373
    - #1374
        - #1375 ◀
            - #1376

PR #1376, opened 2026-06-08
fix(cache): coordinate a compressed request mid-chunking, not 404

kalbasit:
## Summary

- `GetNar` short-circuited to chunk-based serving for an actively-chunking NAR (via `isServable`), 404-ing a compressed (`.nar.xz`) request and skipping download coordination entirely so the cross-pod waiter never recorded a staging request.
- Gate the short-circuit: an actively-chunking NAR (`isServable && !hasFinishedNar`) with a compressed request now falls through to `prePullNar` → coordination. Uncompressed progressive serving and finished NARs are unaffected.

> **Status — IMPORTANT: prerequisite, not a complete fix.** Instrumented debugging shows the chunking-window e2e is still red due to a deeper **lock-lifecycle design mismatch**: the holder releases the `download:nar` lock after the decompress while CDC chunking continues in the background, so the cross-pod reader *acquires* the lock instead of *contending* — bypassing the staging-request path — and hits a **third** `isServable` site (the `coordinateDownload` post-lock check). Naively fixing that site would cause a re-download + double-chunk. Closing #1289 needs a holistic redesign of staging activation under CDC, not another patch; the full root cause is documented for a follow-up.

## Test plan

- [x] `task fmt` / `task lint` / `task test` (race) pass — no regression
- [x] Multi-process e2e: the loser now reaches coordination (chunking-window still red pending the redesign)

### DIFF
--- a/openspec/changes/archive/2026-06-08-fix-cdc-getnar-chunk-route/.openspec.yaml
+++ b/openspec/changes/archive/2026-06-08-fix-cdc-getnar-chunk-route/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-08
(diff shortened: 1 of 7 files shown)


kalbasit:

- #1373
    - #1374
        - #1375
            - #1376 ◀

PR #1379, opened 2026-06-09
fix(cache): hold the NAR download lock through the eager-CDC chunking window

### LINKED ISSUE #1289: CDC: serve whole NAR from shared storage during chunking (+ GC after grace) instead of progressive chunk stream

kalbasit:
## Summary

- Hold the per-hash NAR **download lock through the eager-CDC chunking window** (release on `ds.done`, not `ds.stored` which closes at chunk *start*). A cross-pod reader arriving mid-chunking now **contends** and engages in-flight staging via the already-tested download-window contention path, instead of acquiring the now-free lock and bypassing staging.
- For an **uncompressed** (`.nar`) cross-pod request during chunking, the complete NAR is served from in-flight staging — more robust than the fragile progressive-chunk reassembly.
- For a **compressed** (`.nar.xz`) request during chunking — which **cannot** be reconstructed from the decompressed eager-CDC in-flight bytes (ncps has no NAR compressor, and a re-compressed file would not match the narinfo `FileHash`/`FileSize`) — `serveNarFromStaging` now returns not-found so the client falls back to an upstream that has the original compressed file, rather than being served the uncompressed bytes mislabeled as `xz`.
- Non-CDC and lazy-CDC download-lock timing is unchanged; holder death during chunking is still recovered via the lock TTL. No new flags, schema, or migrations — behavior-only, behind the existing (default-off) inflight-staging flag.

The producer + temp file already survive the chunking window (`waitForStagingRequest` polls until `ds.done`; the temp is removed only after `ds.cdcWg`), so no producer-liveness code was needed — verified by test rather than added.

## Scope / honest caveat

This is the **lock-lifetime + safe-fallback** fix. The cross-pod "served from staging during chunking" behavior is **unit-verified** (lock held through chunking; producer stages a mid-chunk request; compressed→fallback) but has **no green end-to-end run** yet: the contention e2e's chunking-window expectation (serve compressed from staging) is itself unachievable for eager CDC and needs redesigning. The **root fix** — advertise narinfo `none` for CDC NARs so `.nar.xz` is never requested (which also subsumes a pre-existing same-pod corruption of the same shape) — is **deferred** to a follow-up. So this PR does **not** fully close #1289 on its own; it removes the regression path and the corruption, and serves the uncompressed cross-pod case from staging.

Refs #1289.

## Test plan

- [x] \`task fmt\`, \`task lint\` (0 issues), unit \`task test\` green under \`-race\`
- [x] New unit tests: lock held through chunking (red→green); producer stages a request recorded mid-chunking; compressed-request-from-uncompressed-staging → fallback
- [ ] Cross-pod chunking-window e2e — deferred (needs redesign to assert the uncompressed staging path / compressed→fallback)


kalbasit:

- #1379 ◀

PR #1382, opened 2026-06-09
feat: verify upstream signatures on narinfo ingestion

### LINKED ISSUE #1269: feat: verify upstream signatures on narinfo ingestion
## Problem

ncps currently appends its own signature to narinfos on ingestion (both from upstream fetch and client PUT) but does **not verify** the incoming upstream signatures before storing or serving them.

This means ncps will happily cache and re-serve a narinfo with a forged or invalid upstream signature, adding only its own signature on top. A client trusting ncps's key will accept the entry, never knowing the upstream signature was bad.

## Expected behavior

On ingestion, ncps should verify that at least one signature on the narinfo is valid against a configured set of trusted public keys (e.g. the upstream cache's public key configured via `--upstream-cache-public-key`). Entries that carry no valid trusted signature should be rejected.

This is analogous to how `nix` itself validates signatures before substituting from a cache.

## Context

Discovered while designing build-trace-v2 support (the new `build_trace_signatures` storage will follow the same pattern as `narinfo_signatures`). Both should eventually verify on ingestion.

## References

- narinfo signing: `pkg/cache/cache.go` → `signNarInfo()`
- narinfo_signatures table: `ent/schema/`

mvanhorn:
## Summary

Adds optional trusted-signature verification to the client `PutNarInfo` ingestion path. Fixes #1269.

## Why this matters

ncps signs narinfos on ingestion (both upstream-fetch and client `PUT`) and re-serves them under its own key. The upstream-fetch path already verifies incoming signatures against the configured trusted public keys (`signature.VerifyFirst` in `pkg/cache/upstream/cache.go`). The client `PutNarInfo` path, however, parsed, signed, and stored uploads with **zero** signature verification. A client could upload a narinfo carrying a forged or absent upstream signature; ncps would sign it and serve it, and downstream clients trusting ncps's key would accept it without ever knowing the original signature was invalid. nix itself validates signatures before substituting from a cache, so this closes that gap for the unverified `PUT` path.

## Changes

- `pkg/cache/upstream/cache.go`: new `PublicKeys()` accessor exposing the parsed trusted keys (previously unexported with no getter).
- `pkg/cache/cache.go`: new `requireTrustedSignature` field + `SetCacheRequireTrustedSignature` setter (mirrors the existing `shouldSignNarinfo` pattern), a `verifyNarInfoTrusted` helper that aggregates trusted keys across configured upstream caches and runs `signature.VerifyFirst`, and a call to it inside `PutNarInfo` immediately after parse and before signing. Adds the `ErrUntrustedNarInfo` sentinel.
- `pkg/ncps/serve.go`: new `--cache-require-trusted-signature` flag (default off) wired to the setter.

Behavior is unchanged when the flag is off (the default), preserving backward compatibility. When the flag is on, verification is **fail-closed**: an upload with no signature validating against the trusted upstream keys is rejected, and rejected when zero trusted keys are configured (matching nix's posture, so operators cannot accidentally run a no-op verifier).

This is a partial scope of the issue: it targets the unverified client `PUT` ingestion path. The upstream-fetch path already verifies, so it needs only the operator toggle, not new verification logic.

## Testing

- New `PutNarInfoRequireTrustedSignature` test covering: disabled (backward compat accept), fail-closed with no keys configured, untrusted-key rejection (not persisted), and trusted-key success (persisted).
- `go build`, `go vet`, `gofmt -l` clean on affected packages; full SQLite cache suite passes.

AI was used for assistance.

### DIFF
--- a/charts/ncps/templates/configmap.yaml
+++ b/charts/ncps/templates/configmap.yaml
@@ -89,6 +89,7 @@ data:
       {{- end }}
 
       sign-narinfo: {{ ne .Values.config.signing.enabled false }}
+      require-trusted-signature: {{ eq .Values.config.signing.requireTrustedSignature true }}
 
       upstream:
         urls:
--- a/charts/ncps/tests/configmap_test.yaml
+++ b/charts/ncps/tests/configmap_test.yaml
@@ -269,6 +269,23 @@ tests:
           path: data["config.yaml"]
           pattern: "sign-narinfo: true"
 
+  - it: should default require-trusted-signature to false
+    set:
+      config.hostname: test.example.com
+    asserts:
+      - matchRegex:
+          path: data["config.yaml"]
+          pattern: "require-trusted-signature: false"
+
+  - it: should enable require-trusted-signature when set
+    set:
+      config.hostname: test.example.com
+      config.signing.requireTrustedSignature: true
+    asserts:
+      - matchRegex:
+          path: data["config.yaml"]
+          pattern: "require-trusted-signature: true"
+
   - it: should configure upstream caches
     set:
       config.hostname: test.example.com
--- a/charts/ncps/values.schema.json
+++ b/charts/ncps/values.schema.json
@@ -113,6 +113,10 @@
             "enabled": {
               "type": "boolean"
             },
+            "requireTrustedSignature": {
+              "type": "boolean",
+              "description": "Reject PUT-uploaded narinfos lacking a signature trusted by the configured upstream public keys (fail-closed)"
+            },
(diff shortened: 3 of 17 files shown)


kalbasit (approved):

kalbasit:
Thank you @mvanhorn! I pushed docs/helm and openspec changes to your branch and it's looking great now!

mvanhorn:
Appreciate the merge here too, @kalbasit. Verifying upstream signatures at narinfo ingestion is a solid trust boundary.

PR #1383, opened 2026-06-09
feat(e2e): multi-scenario runs + nightly e2e workflow

kalbasit:
## Summary

The unified e2e harness ran exactly one scenario per invocation and had **no automated coverage at all**. This adds multi-scenario selection, the harness's first unit-test net, and a nightly workflow.

- **Multi-scenario selection** — `--scenario` is now repeatable and comma-separated, and a new `--all` runs every catalog scenario supporting the chosen mode. `runner.run_scenarios` reports per-scenario PASS/FAIL/SKIP with an aggregate summary and exits non-zero iff any scenario FAILED (a SKIP alone never fails the run). Single `--scenario <name>` is unchanged.
- **First test net** — fast offline pytest coverage of the CLI/runner logic, wired as the `e2e-harness-unit` flake check and a `task test:e2e:unit` target (the harness had no tests before).
- **Nightly workflow** — `.github/workflows/e2e-nightly.yml` runs the full catalog as a `local`+`kubernetes` matrix (`fail-fast: false`) at 04:00 UTC, plus manual `workflow_dispatch`. A `gate` job records the last successfully-tested `main` SHA in an `actions/cache` key and skips when `main` hasn't advanced, so the same commit is never e2e-tested two nights running. The SHA is recorded only after a fully successful run; manual dispatch bypasses the gate.

## Deferred (follow-up change)

Lifting the `cdc-lifecycle` / `staging-contention` local-only pins onto kubernetes is **carved out**: it needs substantial new k8s plumbing (per-pod port-forward, in-cluster per-dialect DB access, a CDC enable/disable ConfigMap toggle) and long Kind verification cycles. Until then the nightly kubernetes leg SKIPs those two scenarios. See the archived change's `design.md` D2.

## Test plan

- [x] `task fmt`, `task lint`, `task test` (Go, race) all green
- [x] `nix build .#checks.<system>.e2e-harness-unit` green (15 pytest cases)
- [x] `actionlint` clean on the nightly workflow
- [x] `nix run .#e2e -- --list` works via the wrapper
- [ ] First live nightly proves the cache-keyed dedup / matrix / record wiring (can't fire a scheduled workflow pre-merge)

OpenSpec change `e2e-multi-scenario-nightly` archived; main specs `unified-e2e-harness` + new `e2e-nightly-ci` updated.

### DIFF
--- a/.github/workflows/e2e-nightly.yml
+++ b/.github/workflows/e2e-nightly.yml
@@ -1,0 +1,90 @@
+name: e2e-nightly
+# Nightly run of the full unified e2e catalog against both modes. This is the
+# scheduled home for the harness (it stays out of the per-PR `nix flake check`).
+# A `gate` job records the last successfully-tested `main` SHA in an
+# `actions/cache` entry and short-circuits when `main` has not advanced, so the
+# same commit is never e2e-tested on two consecutive nights. A manual
+# `workflow_dispatch` bypasses the gate and forces a run against current `main`.
+on:
+  schedule:
+    - cron: "0 4 * * *" # 04:00 UTC nightly
+  workflow_dispatch:
+permissions:
+  contents: read
+concurrency:
+  group: e2e-nightly
+  cancel-in-progress: false
+jobs:
+  gate:
+    runs-on: ubuntu-24.04
+    outputs:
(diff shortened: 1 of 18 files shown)


kalbasit:

- #1383 ◀

PR #1384, opened 2026-06-09
feat(e2e): run cdc-lifecycle on kubernetes via a Deployment adapter

kalbasit:
## Summary

Adds `KubernetesDeployment` — a `Deployment`-protocol implementation that runs the unified e2e **phase drivers unchanged on Kind** — and routes the `cdc-lifecycle` scenario through it, so it is no longer `local`-only. The runner gates on scenario **name**, so the multi-replica `ha-s3-postgres-cdc-lifecycle` permutation keeps its `NCPSTester` topology checks.

Seams the kubernetes side previously lacked, now implemented:
- **Per-pod addressing** — one `kubectl port-forward pod/<n>` per replica, plus a run.py-shaped `state.json` so `seed_cache` builds through the cluster ncps.
- **CDC enable/lazy toggle** — `helm upgrade --set config.cdc.enabled=… --set config.cdc.lazyChunkingEnabled=…` + `kubectl rollout restart` (the previous tester could only *disable* CDC).
- **`run_subcommand`** — execs the **shell-less** ncps binary directly; during drain (ncps scaled to 0) runs `migrate-chunks-to-nar`/`fsck` in a one-shot pod cloned from the resolved pod spec (preserves the StatefulSet PVC).
- **In-cluster `db()`** — postgres/mysql via port-forward; **sqlite via a `kubectl debug` sidecar** sharing the ncps PID namespace, reading the live DB at `/proc/1/root` (the image is shell-less). During the drain window sqlite is read from a transient pod mounting the released PVC.

### Verified live on Kind
`nix run .#e2e -- --mode kubernetes --scenario cdc-lifecycle` → **PASS, 41/41 checks** (baseline → eager → lazy → drain → restart → fsck).

### Scope decision: `staging-contention` stays local-only
Its adapter seams all work and serving is byte-correct cross-pod, but in-flight staging **activation** is a single-shot timing event that `kubectl port-forward` latency jitter de-synchronizes (the lock-holder caches the NAR before cross-pod waiters contend), so activation can't be reliably forced on Kind. The pin now documents that and the scenario **SKIPs** under `--mode kubernetes`. The adapter is ready to lift it later if the race is made deterministic.

No ncps production code, schema, or migration is touched. Offline unit coverage for the adapter + routing runs in `checks.e2e-harness-unit`.

## Test plan

- [x] `task fmt` / `task lint` (Go: 0 issues)
- [x] `task test` (Go unit, race) green
- [x] `checks.e2e-harness-unit` flake check (34 offline adapter/runner tests)
- [x] `openspec validate --strict`
- [x] Live Kind: `cdc-lifecycle [kubernetes]` 41/41 PASS
- [ ] Nightly `--mode kubernetes --all` exercises it (post-merge; staging-contention SKIPs)


kalbasit:

- #1384 ◀

PR #1387, opened 2026-06-10
fix(e2e): make the e2e-nightly workflow reliable on both modes

kalbasit:
## Summary

The first `e2e-nightly` run failed on both legs. Investigation turned up several independent issues — most surfaced one after another only once CI could get far enough to hit the next one. This makes both legs reliably green. **The harness changes only; no ncps production code is touched.**

### Local leg (was 0/9 passed)
- **Start backing services once per run.** `runner._run_local` started *and* tore down deps per scenario; each backend boots cold into an ephemeral `mktemp` dir, so every scenario paid a full cold boot and raced the readiness timeout. A multi-scenario local run now starts them once and tears down once.
- **Cold-boot-tolerant readiness + diagnostics.** Raise the readiness timeout 120s → 300s, and on timeout dump process-compose state + per-process logs (this is what surfaced every subsequent layer).
- **MariaDB starts on a hosted runner.** `mariadb-install-db` read the runner's pre-installed MySQL `/etc/mysql/my.cnf` (which sets `user=mysql` and the MySQL-only `mysqlx-bind-address`), tried to chown the datadir to an unwritable user and then aborted on the unknown variable, so 3306 never bound. Run MariaDB with `--no-defaults`.
- **ncps actually launches in CI.** `dev-scripts/run.py` launches each replica via `watchexec` + `direnv exec` + `go run` (cgo needs gcc), so the local leg needs the dev-shell toolchain. The nightly ran bare `nix run .#e2e` outside the dev shell, and `direnv` wasn't even in the flake. The local leg now runs inside a **CI-only `ci-e2e-local` devShell** (the default dev shell + `direnv`), so the default devShell / docker-dev image / a developer's host `direnv` are all untouched.

### Kubernetes leg (was 13/15; flaky)
- **Tolerate a transient post-deploy 5xx.** The HTTP-endpoint validation did a single, no-retry narinfo/NAR GET; a transient warm-up 5xx failed the scenario. Added a small dependency-injected retry helper.
- **Consistent SQLite snapshot.** The DB probe copied `ncps.db` + `-wal` + `-shm` as separate files; a WAL checkpoint between the copies produced a torn snapshot missing `nar_files` (while HTTP served narinfo from that very table). Copy `-wal` first and the main DB last, omit the volatile `-shm`, and retry.

### Workflow
- `workflow_dispatch` gains an optional `ref` input (default `main`) so the nightly can be run against a branch: `gh workflow run e2e-nightly.yml --ref <branch> -f ref=<branch>`.

## Test plan

- [x] `task fmt` / `task lint` clean; 54 harness unit tests pass (`e2e-harness-unit`), including new `test_http_retry`, `test_deps`, and a start-once `test_runner` case.
- [x] Local validated in a stripped env (`env -i`, host tools removed) via the exact CI invocation — toolchain resolves from the nix store and scenarios pass.
- [x] **Full `e2e-nightly` run against the branch is green on both legs** (`--mode local` and `--mode kubernetes`).
- [x] OpenSpec change `fix-e2e-nightly-failures` archived; `unified-e2e-harness` spec updated.

### DIFF
--- a/.github/workflows/e2e-nightly.yml
+++ b/.github/workflows/e2e-nightly.yml
@@ -77,7 +77,19 @@ jobs:
           name: ncps
           authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
       - name: Run e2e catalog (${{ matrix.mode }})
-        run: nix run .#e2e -- --mode ${{ matrix.mode }} --all
+        run: |
+          if [ "${{ matrix.mode }}" = "local" ]; then
+            # Local mode drives dev-scripts/run.py, which launches each ncps
+            # replica via watchexec + `direnv exec` + `go run` and therefore
+            # needs the dev-shell toolchain. Run it inside `nix develop` (which
+            # now bundles direnv) with the repo's .envrc allowed so `direnv
+            # exec` can load the flake env. The kubernetes leg is self-contained
+            # in the `nix run .#e2e` wrapper (kubectl/helm/kind) and needs none
+            # of this.
+            nix develop --command bash -c 'direnv allow . && nix run .#e2e -- --mode local --all'
+          else
+            nix run .#e2e -- --mode kubernetes --all
+          fi
   record:
     needs: [gate, e2e]
     # Only record the tested SHA after a fully successful matrix run, so a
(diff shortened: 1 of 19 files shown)


kalbasit:

- #1387 ◀

PR #1389, opened 2026-06-10
feat(dev): round-robin HA proxy on :8500 in run.py

kalbasit:
## Summary

- In HA mode `dev-scripts/run.py` spawns N ncps instances on ports `8501..`, but a Nix client can only point at a single substituter URL. This adds a stdlib round-robin reverse proxy on `127.0.0.1:8500` that fronts all active instances so HA can be exercised from a real Nix client against one stable endpoint.
- Auto-started only when `--replicas > 1`; single-instance runs are unaffected.
- Rotates each request across instances (narinfo → NAR land on different replicas), streams request/response bodies with a bounded buffer (large NARs don't scale proxy memory), strips hop-by-hop headers and pins `Host` to the backend.
- A dead backend yields `502` without killing the proxy thread; a bind failure on `:8500` fails fast after tearing down the just-started instances.
- Lifecycle tied to `run.py`: stopped (port released) in `cleanup()` on SIGINT/SIGTERM; endpoint recorded in `var/ncps/state.json` for test-driver discovery.
- New capability spec `dev-ha-proxy` (proposal/design/specs/tasks archived under `openspec/changes/archive/2026-06-10-run-py-ha-proxy/`).

## Test plan

- [x] `task fmt` clean, `task lint` 0 issues, `task test` all pass
- [x] `dev-scripts/test_run_proxy.py` (8 tests): round-robin + wrap-around, status/header/body passthrough, rotation across two backends, PUT body streaming, dead-backend 502, lifecycle port release, state.json proxy field present/absent
- [x] Live 2-instance smoke (redis locker + postgres): proxy served `nix-cache-info` 200, requests split 4/3 across `:8501`/`:8502`, `state.json` advertised the proxy, SIGTERM cleanup released `:8500` and removed `state.json`

### DIFF
--- a/dev-scripts/run.py
+++ b/dev-scripts/run.py
@@ -2,6 +2,8 @@
 
 import argparse
 import gzip
+import http.client
+import http.server
 import json
 import os
 import shlex
@@ -9,6 +11,7 @@ import shutil
 import signal
 import subprocess
 import sys
+import threading
 import time
 from urllib.parse import urlparse
 
@@ -38,6 +41,7 @@ DB_CONFIG = {
 
 REDIS_ADDR = "127.0.0.1:6379"
 BASE_PORT = 8501
+PROXY_PORT = 8500
 PPROF_BASE_PORT = 7501
 
 # Colors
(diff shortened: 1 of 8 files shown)


kalbasit:

- #1389 ◀

PR #1390, opened 2026-06-10
fix(dev): forward chunked request bodies in HA proxy

kalbasit:
## Summary

The dev HA round-robin reverse proxy in `dev-scripts/run.py` (PR #1389) silently dropped the body of any request sent with `Transfer-Encoding: chunked`. It only streamed the request body when a `Content-Length` header was present; chunked uploads (no `Content-Length`, and `Transfer-Encoding` stripped as hop-by-hop) skipped the body loop entirely, so the backend received a **bodiless PUT** and the unconsumed client body reset the connection — surfacing as `curl error 56: Connection reset by peer` and aborting the whole `nix copy` closure upload.

`nix copy` uploads NARs with chunked encoding because the on-the-fly-compressed `.nar.zst` size is unknown ahead of time, so large closures (e.g. a `nixos-system`) failed while small `Content-Length` NARs succeeded.

### Diagnosis (proven)
Driving the real `run.py` proxy against a capture backend:

| Framing | Client sent | Backend received |
|---|---|---|
| `Content-Length` | 200000 | **200000** ✅ |
| `Transfer-Encoding: chunked` | 200000 | **0** ❌ |

### Fix (`dev-scripts/run.py` only — ncps server untouched)
- Detect chunked framing before the header is dropped; decode the client's chunked body and re-encode it chunked to the backend in bounded `PROXY_BUFFER` units (no whole-body buffering — proxy memory still does not scale with payload size).
- Harden the error path: on backend failure, drain the unconsumed body and force the connection closed so the client gets a real `502` instead of a TCP reset.
- Surface backend connect/forward failures via the log (previously `log_message` was a no-op), so dev upload failures are diagnosable.
- Sync the `dev-ha-proxy` spec (chunked + Content-Length forwarding, 502-not-reset, error visibility) and archive the OpenSpec change.

## Test plan

- [x] New regression test drives the real `run.py` proxy with both `Content-Length` and `Transfer-Encoding: chunked` PUTs; asserts full body reaches the backend in both, and unreachable backend → `502` (not reset).
- [x] Wired `dev-scripts/run.py` into the `e2e-harness-unit` flake check so the test runs in CI (`nix build .#checks.x86_64-linux.e2e-harness-unit` → 43 passed).
- [x] `task fmt` clean · `task lint` 0 issues · `task test` all Go packages pass.

### DIFF
--- a/dev-scripts/run.py
+++ b/dev-scripts/run.py
@@ -384,7 +384,19 @@ def make_proxy_handler(backends):
             """Drain the unconsumed body, send an error, and close the
             connection so the client observes `code` rather than a TCP reset."""
             self.close_connection = True
-            self._drain_request_body(cl_remaining, is_chunked)
+            # Time-bound the drain: a client that stalls mid-upload must not
+            # block us in rfile.read()/readline() and prevent the 502. On
+            # timeout the socket raises (OSError), which _drain_request_body
+            # swallows, stopping the drain.
+            prev_timeout = self.connection.gettimeout()
+            try:
+                self.connection.settimeout(0.25)
+                self._drain_request_body(cl_remaining, is_chunked)
+            finally:
+                try:
+                    self.connection.settimeout(prev_timeout)
+                except OSError:
+                    pass
             try:
                 self.send_error(code, message)
             except OSError:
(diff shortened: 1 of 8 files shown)


kalbasit:

- #1390 ◀

PR #1391, opened 2026-06-10
fix(dev): HTTP/1.1 keep-alive in HA proxy (run.py)

kalbasit:
## Summary

A **second, distinct** dev-harness bug from the chunked-body fix in #1390 (which is necessary and stays). After #1390, `nix copy` of a large closure (e.g. a ~9.35 GB `nixos-system`) to the `:8500` HA proxy **still** intermittently fails with `curl error 56: Connection reset by peer`, aborting the whole copy on the first reset.

**Root cause:** the proxy's `ProxyHandler` left `protocol_version` at the stdlib `BaseHTTPRequestHandler` default of **HTTP/1.0**, so it force-closed every client connection after one request. `nix copy` pushes hundreds of objects over a small pool of HTTP/1.1 keep-alive connections; because the proxy refused keep-alive, nix was forced to open a brand-new TCP connection per object — thousands for a multi-GB closure. Under that churn (amplified by the default listen backlog of 5 and a ~1 s-per-PUT stall from unanswered `Expect: 100-continue`), the accept loop stalls and intermittently nix's upload connection is reset **before the proxy ever handles it** (`curl error 56`).

**Reproduced & fix-validated** on a clean isolated ncps behind the real proxy: the unpatched proxy dies at PUT #2; the patched proxy completes **3,488 PUTs with 0 resets**. `curl -v` confirms `HTTP/1.1 200` + "Connection left intact" (reuse).

## What changed (`dev-scripts/run.py` proxy only; ncps unchanged)

- `ProxyHandler.protocol_version = "HTTP/1.1"` — persistent, reused client connections; also makes `BaseHTTPRequestHandler` auto-answer `Expect: 100-continue`, removing the per-PUT stall.
- Response framing for safe reuse: preserve `Content-Length`; treat `204`/`304`/`HEAD`/`1xx` as bodiless; send `Connection: close` when the body length is unknown (e.g. backend chunked, stripped as hop-by-hop) so the client reads to EOF instead of mis-framing the next response.
- Raise the listener backlog (`request_queue_size`) above the stdlib default of 5, set before the socket is activated.
- All #1390 hardening preserved (chunked request-body forwarding, RFC 7230 §3.3.3, bounded `_fail` drain, error logging).

## Test plan

- [x] `task fmt` (0 changed), `task lint` (0 issues), `task test` (all Go packages ok)
- [x] New keep-alive / framing / `Expect: 100-continue` regression tests drive the real `run.py` proxy via `NCPS_RUN_PY`; confirmed RED on HTTP/1.0, GREEN after the fix
- [x] `e2e-harness-unit` nix check green (52 passed; +4 new tests)
- [x] End-to-end: full `nix copy` of the 9.35 GB closure through the patched proxy — 3,488 PUTs, 0 resets

### DIFF
--- a/dev-scripts/run.py
+++ b/dev-scripts/run.py
@@ -527,7 +527,13 @@ def make_proxy_handler(backends):
                     or self.command == "HEAD"
                     or 100 <= resp.status < 200
                 )
-                if not has_length and not bodiless:
+                # Advertise the close when we will close for any reason — an
+                # unframed body (above), or the connection was already marked to
+                # close (e.g. the client sent `Connection: close`, or it is an
+                # HTTP/1.0 client). RFC 7230 §6.3 requires the server to send
+                # `Connection: close` in that case so a keep-alive client does
+                # not assume the connection persists and reuse it.
+                if self.close_connection or (not has_length and not bodiless):
                     self.send_header("Connection", "close")
                     self.close_connection = True
                 self.end_headers()
--- a/dev-scripts/run.py
+++ b/dev-scripts/run.py
@@ -319,6 +319,18 @@ def make_proxy_handler(backends):
     lock = threading.Lock()
 
     class ProxyHandler(http.server.BaseHTTPRequestHandler):
+        # Speak HTTP/1.1 so client connections are persistent and reused across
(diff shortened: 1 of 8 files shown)


kalbasit:

- #1391 ◀

PR #1393, opened 2026-06-10
fix(cache): serve uncompressed NAR from any stored whole-file compression

### LINKED ISSUE #1329: HTTP 500 "invalid nar hash" when an upstream serves an opaque (non-hash) NAR URL — e.g. cachix UUID-named NARs
## Summary

When ncps proxies a narinfo from an upstream whose `URL:` field is **not** a Nix-hash-named NAR (e.g. cachix now serves `nar/<uuid>.nar.zst`), `ParseURL` rejects it and the request fails with **HTTP 500 `invalid nar hash`**. Because nix treats a 500 as a hard error (worse than a 404, which would fall through to the next substituter), this breaks substitution for every affected path — observed "firing on tons of nars" once a fleet uses ncps as its sole substituter in front of cachix-backed caches.

## Reproduction

```
$ curl -s https://nix-community.cachix.org/9scxyz46ynlawfzm22s6vl603m9syz3w.narinfo | grep -E '^URL:|^Compression:|^NarHash:'
URL: nar/d0c36585-67ac-4e1e-8747-3af0cbc09b90.nar.zst   # <-- UUIDv4, not a nix hash
Compression: zstd
NarHash: sha256:13ishym4qd67vr233nkw6fv8i9x6nzpqmh9i4b8wz1jsgmsy1isp
```

Point an ncps instance at that cache as an upstream and request the narinfo through ncps:

```json
{"level":"error","cmd":"serve","request_uri":"/9scxyz46ynlawfzm22s6vl603m9syz3w.narinfo",
 "error":"invalid nar hash","nar_url":"nar/d0c36585-67ac-4e1e-8747-3af0cbc09b90.nar.zst",
 "message":"error parsing the nar URL"}
{"level":"error","...,"error":"invalid nar hash","message":"error fetching the narinfo"}
{"level":"info","...,"status":500,"bytes":17,"message":"handled request"}
```

A contrasting cache.nixos.org path (`nar/<52-char-base32>.nar.xz`) serves **200** fine.

## Root cause

`pkg/nar/url.go` → `ParseURL` takes everything before `.nar` in the filename as the NAR *hash* and validates it via `ValidateHash` (`pkg/nar/hash.go`):

```go
const NormalizedHashPattern = `(?:[0-9a-df-np-sv-z]{52}|[0-9a-f]{64})`
```

A UUID (`d0c36585-67ac-4e1e-8747-3af0cbc09b90` — hyphens, out-of-alphabet `o/t/u`, wrong length) matches neither the 52-char base32 nor the 64-char hex form, so `ParseURL` returns `ErrInvalidHash` and the whole narinfo fetch aborts with 500.

ncps reuses this parsed hash as its own storage key (`URL.ToFilePath` → `FilePath(hash, ext)`), which is why it insists the upstream URL encode a hash.

## Why the upstream is not at fault

The Nix binary-cache protocol treats the narinfo `URL:` field as an **opaque path relative to the cache root** — there is no requirement that it be hash-named. Clients are expected to read the authoritative URL from the narinfo, not reconstruct/assume it:

- Hash-naming (`nar/<filehash>.nar.xz`) is a convention of `nix copy` / nix-serve / cache.nixos.org, not part of the spec.
- [NixOS/nix#4464](https://github.com/NixOS/nix/pull/4464) tried to *standardize* NAR-hash-based URL naming and was **merged then reverted**, deliberately keeping the field flexible.
- Cachix exposes the NAR under an opaque multipart object key (a UUIDv4 — RFC 9562). This is spec-compliant.

So opaque NAR URLs are valid upstream behavior that any pull-through proxy should tolerate. The fix belongs in ncps.

## Affected versions

The NAR-URL parser is **byte-for-byte identical** between `v0.9.4` (current nixpkgs stable) and `v0.10.0-rc9`/`main`, so every released and pre-released version is affected. Upgrading does not help.

### AI Transparency
LLMs were used in the development of this issue.

### LINKED ISSUE #1392: NAR serve fails ("does not exist"/decode error) on narinfo↔nar_file compression desync
## Summary

Serving a NAR fails with `does not exist in binary cache` (or a decode error) when the **narinfo's advertised compression differs from how the NAR is physically stored** — even though the bytes are present locally. This is the `nar/<hash>.nar(.zst)` "does not exist" failure that aborts `nix` builds intermittently.

This is **distinct from the closed #1329** (opaque/UUID NAR URL parsing). It is a serve-path compression mismatch.

## Evidence (production)

Joining `narinfos` ⇆ `nar_files` by stored compression on a live cache:

```
narinfo │ stored │ count
none    │ none   │ 91716   healthy
zstd    │ zstd   │  1657   healthy
xz      │ xz     │   628   healthy
none    │ xz     │   110   ❌ request .nar → 404 "does not exist"
none    │ zstd   │    90   ❌ request .nar → 404 "does not exist"
xz      │ none   │    53   ❌ request .nar.xz → served raw → "Lzma: No progress is possible"
xz      │ zstd   │     6   ❌ wrong codec → decode error
```

~259 store paths have a narinfo whose advertised `Compression`/`URL` does not match the stored NAR. Each reported client failure (`nar/<hash>.nar does not exist`) maps to a row where the narinfo advertises `none` while the NAR is stored whole-file as `xz`.

## Root cause

The serve path resolves a NAR by an (approximately) exact `(hash, compression)` match and only ever transparently decompresses a stored **`.nar.zst`** for a `Compression:none` request (`getNarFromStore`/`statNarInStore` in `pkg/cache/cache.go`). A NAR stored as `.nar.xz` while the narinfo advertises `none` is therefore reported absent → 404 / re-download, despite the bytes being on disk.

The drift is generated by CDC (which normalizes narinfos to `Compression: none`) and remains as residue after CDC is disabled and NARs are re-downloaded at the upstream's compression. It reproduces both:

- after converting a cache from CDC **on → off** (narinfo `none`, stored `xz`), and
- on a steady-state CDC deployment (independently reported: `<hash>.nar.zst does not exist`, served from uncompressed chunks).

so it is **CDC-state-independent**.

## Reproduction (unit)

`pkg/cache`: store a NAR as a whole-file `.nar.xz`, then request it uncompressed (`Compression:none`) with no upstream configured. The bytes are present (`HasNarInStore` is true for the xz form) but `GetNar(none)` returns `not found`.

## Fix (this change)

Generalize the existing zstd transparent-decompress convention to **any stored whole-file compression** (adds `xz`) in both the existence check (`statNarInStore`) and the serve path (`getNarFromStore`): resolve the NAR by hash, decompress the stored whole file to satisfy a `none` request, and touch the row matching the bytes actually served. Covers the dominant direction (`none` advertised / compressed stored).

## Follow-ups (separate)

- The inverse direction (narinfo advertises `xz`/`zstd` while stored uncompressed or as chunks) needs serve-time **re-compression** and is deferred.
- A one-shot data repair to reconcile existing drifted narinfo/nar_file pairs.
- Write-path invariant: keep a narinfo's advertised compression equal to a servable representation so drift cannot accumulate.

kalbasit:
## Summary

Fixes the production `error: file 'nar/<hash>.nar' does not exist in binary cache` (and `Lzma: No progress is possible`) failures that abort `nix` builds even though the NAR bytes are present locally.

**Root cause — narinfo↔nar_file compression desync on the serve path.** A NAR whose narinfo advertises `Compression: none` can be physically stored whole-file as `.nar.xz`. The serve path (`getNarFromStore` / `statNarInStore`) only ever transparently decompressed a stored `.nar.zst` for a `none` request — never `.nar.xz` — so it reported the NAR absent and 404'd / re-downloaded. With the upstream missing or unhealthy the client saw "does not exist in binary cache".

The drift is generated by CDC (which normalizes narinfos to `Compression: none`) and left as residue after CDC is disabled and NARs re-download at the upstream's compression. It is **CDC-state-independent** — reproduced both after converting a cache CDC on→off and on a steady-state CDC deployment.

**Evidence.** DB inspection of the live cache found ~259 drifted narinfo/nar_file pairs (`none|xz` 110, `none|zstd` 90, `xz|none` 53, `xz|zstd` 6); every client-reported failing hash was `narinfo=none` / `stored=xz`.

## What changed

- New `wholeFileServeCompressions()` helper (`[zstd, xz]`, preference order).
- `getNarFromStore`: for a `Compression: none` request, find whichever supported whole-file compression is stored and decompress it transparently via `nar.DecompressReader`; the served stream is labeled `none`; transparent-zstd pass-through stays gated on the stored file being zstd.
- `statNarInStore`: the existence check reports present when any servable whole-file compression is on disk.
- LRU touch follows the bytes actually served — touches the stored compression's `nar_file` row instead of synthesizing a spurious `none` row; healing only when no row exists.

Covers the dominant direction (`none` advertised / compressed stored).

## Scope / follow-ups

Deferred (tracked in #1392): the inverse direction (narinfo advertises `xz`/`zstd` while stored uncompressed or as chunks — needs serve-time re-compression), a one-shot data repair of existing drifted pairs, and a write-path invariant keeping a narinfo's advertised compression equal to a servable representation.

Distinct from the closed #1329 (opaque/UUID nar-url parsing).

Closes #1392

## Test plan

- [x] New proving test `TestServeNoneRequestFromXzWholeFile` (RED→GREEN): store a NAR whole-file as `.nar.xz`, no upstream, request it as `Compression: none` → served decompressed; no spurious `none` row created.
- [x] `TestCDCWholeFileServeReportsServedCompression` and all `TestGetNar*` still pass.
- [x] `task fmt`, `task lint`, `task test` all green.


kalbasit:

- #1393 ◀

kalbasit:
Thanks for the reviews — addressed:

**CodeRabbit (delete/GC symmetry, `PRRT_kwDONV6tZs6IorjU`):** valid catch, fixed in efb0a8e2. `DeleteNar` and `deleteNarBytes` now iterate `wholeFileServeCompressions()` so a drifted `none→xz` NAR's `.nar.xz` blob is deleted rather than leaked (and `DeleteNar` no longer returns `ErrNotFound` for it). Added `TestDeleteNoneDeletesXzWholeFile`.

**Gemini (always touch `storedComp` row, `PRRT_kwDONV6tZs6IooOT`):** not applying — it would regress the canonical case. An uncompressed (`Compression: none`) NAR is physically stored as `.nar.zst` but its `nar_file` row is `compression=none` (with `bytes_stored_at` set); there is **no** `zstd` row. Always touching `storedComp` (= `zstd`) would miss the existing `none` row and trigger a spurious heal on *every* uncompressed serve. The current logic touches the served-representation (`none`) row when it exists and falls back to the stored compression's row only for the desync case (none requested, only an `xz` row exists), which is the correct behavior. Covered by `TestServeNoneRequestFromXzWholeFile` (asserts no spurious `none` row is created).

PR #1394, opened 2026-06-11
fix(cache): serve a zstd request by recompressing CDC chunks

kalbasit:
## Summary

Inverse direction of the merged uncompressed-serve fix (#1393) and the steady-state-CDC manifestation of the narinfo↔nar_file compression desync (#1392). A client whose narinfo advertises `Compression: zstd` requests `/nar/<hash>.nar.zst`, but under CDC the NAR is stored only as uncompressed chunks. `serveNarFromStorageViaPipe` returned `404` ("only available as chunks, cannot serve as zstd"), aborting the build with `does not exist in binary cache`. zstd is the common upstream compression (cachix), so this blocks CDC-on caches.

Recompress on the fly: reassemble the uncompressed chunks via `getNarFromChunks` and stream them through `zstd.NewPooledWriter` into an `io.Pipe` (mirroring the existing none→zstd store path), serving a `zstd`-labeled response. Scoped to zstd, the one streaming compressor ncps vendors; `xz` (no compressor) keeps the 404/fallback and is handled by the data-repair PR stacked on top.

## Test plan

- [x] New `TestServeZstdRequestFromChunks` (RED→GREEN): chunk a NAR under CDC, request it as `zstd`, assert the served stream decompresses to the original (was 404).
- [x] Existing `TestGetNar*` / CDC serve tests pass; `task fmt`/`lint`/`test` green.

Refs #1392

### DIFF
--- a/openspec/changes/archive/2026-06-10-serve-compressed-nar-via-zstd-recompression/.openspec.yaml
+++ b/openspec/changes/archive/2026-06-10-serve-compressed-nar-via-zstd-recompression/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-10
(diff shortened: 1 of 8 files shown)


kalbasit:

- #1394 ◀
    - #1395
        - #1396

kalbasit:
Good catch — fixed in the latest push. Adopted the defer-based teardown, with one improvement over the suggestion: `pipeWriter.CloseWithError(...)` is now *inside* the defer too, so a panic in `io.Copy` also closes the pipe and the consumer never hangs (the suggested version left the pipe close outside the defer). `PooledWriter.Close` is idempotent, so the single close in the defer is sufficient.

PR #1395, opened 2026-06-11
feat(fsck): repair narinfos advertising a non-producible compression

kalbasit:
## Summary

Stacked on #1394. After the serve-side fixes (#1393, #1394) ncps tolerates most narinfo↔nar_file compression drift at serve time by transcoding. The one case it still cannot serve is a narinfo advertising a compression ncps has no compressor for (`xz`) whose backing NAR is stored otherwise (none/zstd/chunks) — ~59 such pairs on the live cache. These pre-existing drifted rows cannot be healed at request time, so they need a one-shot data repair.

Add `repairNarInfoCompressionDesync`, invoked from `repairFsckIssues` under `--repair`: find narinfos advertising `xz` with no matching `xz` nar_file but a backing non-xz nar_file, and rewrite each to the servable `none` form (`URL: nar/<hash>.nar`, `Compression: none`, `FileHash`/`FileSize` cleared), reusing the advertised hash so the resolve-by-hash serve path (#1393) decompresses the stored bytes. Healthy narinfos are untouched; the repair is idempotent.

## Test plan

- [x] New `TestRepairNarInfoCompressionDesync`: a desynced xz-advertised/none-stored narinfo is rewritten to none; a healthy xz/xz pair is untouched; a second run is a no-op.
- [x] `task fmt`/`lint` green; targeted fsck tests pass.

Refs #1392

### DIFF
--- a/openspec/changes/archive/2026-06-10-repair-narinfo-compression-desync/.openspec.yaml
+++ b/openspec/changes/archive/2026-06-10-repair-narinfo-compression-desync/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-10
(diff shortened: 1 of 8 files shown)


kalbasit:

- #1394
    - #1395 ◀
        - #1396

kalbasit:
Done — narrowed the query with the suggested predicate (advertised xz, has a linked nar_file, no linked xz nar_file) so only the desynced rows are loaded. Kept the per-row `repairOneNarInfoCompressionDesync` verification (by advertised hash) as a precise re-check before rewriting.

kalbasit:
Good catch — tightened the backing guard to require a nar_file matching the advertised hash+query (still linked to the narinfo), so a stale/mismatched link can no longer cause a rewrite to a non-servable `none` URL. The advertised xz row is excluded just above, so a match here is a non-xz row the serve path can produce `none` from; and the relink repair runs earlier in the same `fsck --repair` pass, so legitimately-backed narinfos are linked by the time this runs.

PR #1412, opened 2026-06-20
feat(cache): trust uploads via dedicated upload keys

kalbasit:
## Summary

The `cache.require-trusted-signature` PUT gate verified uploaded narinfos against the configured **upstream** public keys. Upstream keys can never match a locally-built store path (unsigned, or signed by the operator's own key), so enabling the gate made it impossible to `nix copy` your own builds — every upload failed with `rejecting untrusted narinfo` (HTTP 500). The gate conflated **pull-trust** (which upstreams ncps mirrors) with **upload-trust** (whose uploads ncps re-signs and serves).

This decouples the two:

- New operator-configured `cache.trusted-upload-keys` (flag `--cache-trusted-upload-key`, env `CACHE_TRUSTED_UPLOAD_KEYS`; nix-format `name:base64`), independent of the upstream keys.
- `require-trusted-signature` now verifies PUT narinfos against the upload keys. It remains the **single** enable switch — only its key source changes.
- Fail-closed is preserved: gate enabled with an empty upload-key set rejects every upload.
- The anti-laundering goal (#1269) holds — only operator-trusted signatures are re-signed and served.

To push self-built paths: sign them (`nix store sign` / `secret-key-files`) and list the matching public key in `trusted-upload-keys`.

**Not breaking:** the PUT signature gate is unreleased (absent from the CHANGELOG), so there are no existing users relying on the old upstream-key behavior.

## Changes

- `pkg/cache/cache.go`: `trustedUploadKeys` field + `SetCacheTrustedUploadKeys`; `verifyNarInfoTrusted` consults upload keys; removed now-unused `trustedPublicKeys`.
- `pkg/ncps/serve.go`: new repeatable flag + `parseTrustedUploadKeys` (fails startup on a malformed key).
- Config/docs/chart: `config.example.yaml`, Configuration Reference, Helm `values.yaml` + `configmap.yaml` + unit tests, CHANGELOG.
- Spec: `narinfo-trusted-signature` updated; change archived under `openspec/changes/archive/`.

## Test plan

- [x] Cache unit tests cover all 5 gate scenarios (incl. upstream-key-but-no-upload-key → reject)
- [x] Serve helper tests (parse valid/empty/malformed keys)
- [x] `helm template` verified: key omitted when empty, rendered when populated
- [ ] Live smoke: `nix copy` with a trusted upload key succeeds; empty list rejects

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,20 @@ project loosely follows [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- **Trusted-signature gate on PUT uploads.** A new
+  `--cache-require-trusted-signature` flag (env `CACHE_REQUIRE_TRUSTED_SIGNATURE`,
+  **off by default**) makes ncps verify client-uploaded (`PUT`) narinfos before
+  re-signing and persisting them, closing the signature-laundering gap. When
+  enabled, an upload is accepted only if it carries a signature trusted by the
+  operator-configured `--cache-trusted-upload-key`s (repeatable, env
+  `CACHE_TRUSTED_UPLOAD_KEYS`; nix-format `name:base64`). Upload-trust is
+  **independent** of the upstream public keys (which govern pull-trust), so
+  operators authorize their own build/signing key to push self-built paths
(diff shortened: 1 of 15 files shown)


kalbasit:

- #1412 ◀

PR #1417, opened 2026-06-23
test(e2e): isolate S3 bucket per kubernetes scenario

kalbasit:
## Summary

Fixes the `e2e-nightly` kubernetes leg, which has failed every scheduled run since 2026-06-11 on the CDC scenarios (`single-s3-postgres-cdc`, `ha-s3-postgres-cdc`, `ha-s3-postgres-cdc-lifecycle`) with `PostgreSQL database is empty (0 chunks)`.

**Root cause (confirmed via local Kind e2e + pod logs + bucket-empty proof):** the kubernetes harness isolates a database per scenario but shared a single Garage bucket (`ncps-bucket`) across every scenario and never cleaned it. Earlier non-CDC scenarios reuse the same test narinfo hashes and leave whole-file `.nar.xz` objects in `store/nar/`. When a CDC scenario then runs with a fresh database against that dirty bucket, ncps (since #1393 treats a stored `.nar.xz` as a present `Compression:none` NAR) sees the residual whole file as already present, **skips the upstream download + eager chunking**, and `BackgroundMigrateNarToChunks` fails with `error fetching nar from store: not found` — so zero chunk rows are created and validation fails. Pre-#1393 the path only checked `.nar.zst`, missed the residual, re-downloaded and eager-chunked (the last green run produced 4 chunks). `cdc-lifecycle` passes because it uses local per-pod storage with no shared residue.

## What changed (test harness only — no ncps / chart / CDC behavior change)

- `harness_config.scenario_bucket_name()` — single source of truth deriving `ncps-<scenario>`.
- `config.nix` — each scenario's generated Helm values use its own bucket.
- `k8s_tests.py` — garage setup creates + grants a bucket per S3 scenario (idempotent); test-config carries the per-scenario bucket.
- `k8s_tests_tester.py` — S3 storage validation reads the per-scenario bucket.

A per-scenario **bucket** (not prefix) isolates the NAR, narinfo, **and** chunk stores at once — the chunk store ignores the configured S3 prefix, so a per-scenario prefix would leave chunks colliding.

## Test plan

- [x] `nix eval` confirms each S3 scenario renders a distinct `ncps-<name>` bucket; local scenarios get none
- [x] New harness unit test (`test_s3_bucket_isolation.py`); `e2e-harness-unit` net green (66 passed)
- [x] `task fmt`, `task lint`, `task test` (0 failures) green
- [x] Live Kind run: `single-s3-postgres` (non-CDC) then all three CDC scenarios → **4/4 pass, 4 chunks each** (cross-scenario isolation proven)
- [ ] `e2e-nightly` `workflow_dispatch` on this branch goes green (kubernetes leg)

### DIFF
--- a/nix/e2e-tests/config.nix
+++ b/nix/e2e-tests/config.nix
@@ -614,6 +614,10 @@ rec {
                       # written by another (mirrors the per-scenario ncps_<name>
                       # database). Must match harness_config.scenario_bucket_name
                       # and the buckets created in k8s_tests.py garage setup.
+                      # Scenario names are kebab-case by the catalog contract, so
+                      # this is already a valid S3 bucket name; kept verbatim (not
+                      # normalized) to stay byte-for-byte in sync with the Python
+                      # side (plain Nix has no toLower builtin).
                       bucket = "ncps-${perm.name}";
                       inherit (s3) endpoint;
                       region = "us-east-1";
--- a/nix/e2e-tests/config.nix
+++ b/nix/e2e-tests/config.nix
@@ -610,7 +610,11 @@ rec {
                     # s3
                     type = "s3";
                     s3 = {
-                      inherit (s3) bucket;
+                      # Per-scenario bucket so no scenario observes objects
+                      # written by another (mirrors the per-scenario ncps_<name>
+                      # database). Must match harness_config.scenario_bucket_name
+                      # and the buckets created in k8s_tests.py garage setup.
+                      bucket = "ncps-${perm.name}";
                       inherit (s3) endpoint;
                       region = "us-east-1";
                       prefix = perm.name;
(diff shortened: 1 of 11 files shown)


kalbasit:

- #1417 ◀

kalbasit:
Addressed review feedback in 4e80be84:

- **config_file default path** + **null-storage guard** — fixed as suggested.
- **Bucket-name normalization (both threads)** — respectfully declined. Scenario names are kebab-case by the catalog contract (the `unified-e2e-harness` spec mandates a stable kebab-case name), so `ncps-<name>` is already a valid S3 bucket name. Also, `builtins.toLower` is not a Nix builtin, so the suggested Nix normalization wouldn't evaluate; normalizing only the Python side would risk the Python and Nix bucket names drifting apart. Both sides are kept verbatim so they stay byte-for-byte identical (verified via `nix eval` vs the Python helper: 0 mismatches across all 11 S3 scenarios). Comments on both sides now document this.

PR #1427, opened 2026-07-01
fix(cache): proxy snix-castore opaque nar URLs

kalbasit:
## Summary

Fixes intermittent `HTTP 500 "invalid nar URL"` when serving `<hash>.narinfo`
for store paths available on `cache.snix.dev`.

**Root cause (confirmed against production logs):** snix serves narinfos whose
`URL:` field is a snix-castore reference — `nar/snix-castore/<blob>?narsize=N`,
`Compression: none`, **no `.nar` extension**. ncps's `nar.ParseUpstreamURL`
required a `.nar` token, so it rejected these with `ErrInvalidURL` and the
narinfo request 500'd. It was intermittent because upstream selection races all
healthy upstreams (first responder wins); a retry that landed on
cache.nixos.org served a conventional URL and succeeded.

**Fix:** follow the Nix HTTP binary-cache protocol, where the narinfo `URL:` is
an opaque path relative to the cache root. This generalizes the existing opaque
(cachix UUID) handling to `.nar`-less URLs:

- `ParseUpstreamURL` recovers a genuine cache-relative `.nar`-less path as
  `Compression: none`, keyed off the narinfo `NarHash`, preserving the full path
  **and query** for the upstream GET (snix returns `400` without `?narsize=N`).
- The initial pull now splits the opaque **download** URL (path + query) from
  ncps's own hash-named **storage/serve** key, so the upstream query never leaks
  into the `nar_file` key (which otherwise 404'd the NAR on a later read).
- New `OpaqueUpstreamRef` / `WithOpaqueUpstreamRef` persist and restore the
  opaque path + query so an evicted NAR re-fetches correctly.
- Strict `ParseURL` is unchanged — ncps still emits only hash-named URLs to
  clients. Conventional and cachix-style URLs are unaffected.

No config, schema, or migration changes.

## Test plan

- [x] `task fmt` (0 changed), `task lint` (0 issues), `task test` (all green)
- [x] `pkg/nar`: parser unit tests (snix parse, query round-trip, strict
      `ParseURL` still rejects, invalid-fallback rejected)
- [x] `pkg/cache`: `TestGetNarInfoSnixCastoreOpaqueURL` — full pull → store →
      serve → evict → re-fetch; asserts `?narsize` reaches the upstream GET and
      never leaks into ncps's storage key; existing cachix opaque test still
      passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,6 +135,23 @@ project loosely follows [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- **snix-castore (and other `.nar`-less opaque) upstream narinfo URLs are now
+  proxied instead of returning `HTTP 500 "invalid nar URL"`.** Upstreams such as
+  `cache.snix.dev` serve narinfos whose `URL:` field is a content-addressed
+  reference like `nar/snix-castore/<blob>?narsize=N` (`Compression: none`, no
+  `.nar` extension). ncps previously rejected these while parsing, so a request
+  for such a `.narinfo` failed with a 500 — intermittently, because upstream
+  selection races and a retry that landed on a conventional cache (e.g.
+  cache.nixos.org) succeeded. Following the Nix HTTP binary-cache protocol (the
+  narinfo `URL:` is an opaque path relative to the cache root), ncps now tolerates
+  `.nar`-less opaque URLs: it fetches the NAR from the original opaque path —
+  preserving the required query string (snix returns `400` without `?narsize=N`) —
+  keys local storage off the narinfo `NarHash`, and re-serves under its own
+  hash-named `nar/<hash>.nar` (`Compression: none`). The opaque path and query are
+  persisted so the NAR is re-fetchable from upstream after local eviction. This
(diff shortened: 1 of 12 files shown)


kalbasit:

- #1427 ◀

kalbasit:
Re: preserving the raw opaque query string (thread on `pkg/nar/url.go`) — resolving as a documented non-change:

- The only known `.nar`-less opaque upstream (snix-castore) uses a single `?narsize=<digits>` parameter, which round-trips through `url.Values.Encode()` byte-for-byte (asserted exactly in `TestParseUpstreamURL` / the opaque-ref round-trip test).
- ncps already handles **all** NAR-URL queries via `url.Values` end-to-end — `parseURLParts` decodes with `url.ParseQuery` and `JoinURL` re-emits with `Query.Encode()` for conventional URLs too. Adding a raw-query field only for opaque URLs would be inconsistent with that and a non-trivial change to the shared `URL` type / `JoinURL`.

So opaque re-fetch is no less faithful than the shipped behavior for every other URL, and correct for the real upstream. If an order-/encoding-sensitive multi-param opaque upstream ever appears, carrying a raw query through the `URL` type would be the right follow-up.

Issue #1470, opened 2026-08-17
rc16 serves raw NAR while narinfo declares zstd when upstream URL ends in .nar


Ramblurr:
On v0.10.0-rc16 with CDC disabled, an Attic upstream returns a narinfo whose URL ends in .nar but its `Compression` is `zstd`.

NCPS keeps `Compression: zstd`, but links the narinfo to a nar_file with `compression=none`.

Nix then receives raw NAR bytes and fails with input compression not recognized.

Shouldnt it be the case that the URL isn't soley used to determine the compression scheme?

I did some looking before making this issue and there seem to be some related-but-not-the-same issues:

- my issue looks like the reverse direction of #1392 and #1393, 
- while #1399 covers in-flight CDC, while CDC is disabled here. 
-  #1329 too somehow

kalbasit:
Thanks for the very precise report — and your instinct that this is "the reverse direction of #1392" is exactly right. I've reproduced it locally and root-caused it. Summary below, plus one question I need answered to know whether I've reproduced *your* exact failure or only half of it.

## What Attic sends

From Attic's own source (`server/src/database/entity/object.rs`, `to_nar_info`):

```rust
url:         format!("nar/{}.nar", self.store_path_hash.as_str()),  // always a bare ".nar"
compression: Compression::from_str(&nar.compression)?,              // zstd
file_hash:   None,  // FIXME
file_size:   None,  // FIXME
```

So an Attic narinfo has two properties that are unusual together:

1. the URL carries **no compression signal at all**, and
2. the filename is the **32-char store path hash**, which is neither a 52-char nix32 nor a 64-char hex NAR hash.

`Compression:` is therefore the *only* statement of how the NAR is encoded.

## Where ncps loses it

You're right that the URL shouldn't be the sole source of truth — that's precisely the bug.

Because the 32-char store path hash fails `ValidateHash`, `nar.ParseUpstreamURL` takes the **opaque** branch (the same lane as cachix's UUID NARs), yielding `Compression: none` from the missing file extension. `pullNarInfo`'s normalization switch then hits this arm:

```go
case narURL.IsOpaque():
    // ... re-serve under ncps's own hash-named URL keyed off the NarHash,
    // preserving compression.
    rewrittenURL := nar.URL{Hash: narURL.Hash, Compression: narURL.Compression}
    narInfo.URL = rewrittenURL.String()
```

[...]

PR #1474, opened 2026-08-19
fix(nar): trust the narinfo Compression header when the URL carries no extension

kalbasit:
## Summary

Fixes the narinfo↔nar_file compression desync reported in #1470.

ncps derived a NAR's compression **exclusively from the file extension on the narinfo `URL:` field**, never reconciling it against the narinfo's own `Compression:` header. Attic states compression only in the header — from `attic/server/src/database/entity/object.rs`:

```rust
url:         format!("nar/{}.nar", self.store_path_hash.as_str()),  // always a bare ".nar"
compression: Compression::from_str(&nar.compression)?,              // zstd
file_hash:   None,  // FIXME
file_size:   None,  // FIXME
```

So ncps read `none` and:

- linked a `Compression: zstd` narinfo to a `compression=none` `nar_file` row,
- re-compressed the already-compressed NAR **a second time** on disk, and
- advertised a URL whose bytes did not match the `Compression:` it re-signed and served.

## What changed

`nar.ParseUpstreamURL` now takes the narinfo's declared compression and applies it when the URL carries no extension. An **explicit extension still wins**, so cache.nixos.org, cachix, Harmonia and nix-serve are untouched. An unrecognised declaration is ignored rather than trusted — `CompressionType` is a bare string type, so an unknown value would otherwise reach `ToFileExtension` and panic.

Resolving the compression breaks the upstream path reconstruction: ncps would request `nar/<hash>.nar.zst` while the upstream serves the extension-less `nar/<hash>.nar`. `opaquePath` is therefore generalized from *"the URL is not hash-named"* to *"the upstream path cannot be rebuilt from Hash+Compression"*. That preserves the original path for the upstream GET, persists it for re-fetch after eviction, and routes both the opaque (Attic, cachix) and conventional hash-named shapes through the existing `IsOpaque` arm of `pullNarInfo`'s normalization switch.

`storeInDatabase` and `putNarInStore` needed no change once they receive the resolved value — the rows agree and the double-compression disappears on its own.

Note the fix could not live in the opaque branch alone: Attic reaches it only because its 32-char store-path hash fails `ValidateHash`. A conforming upstream serving `nar/<52-char-narhash>.nar` with `Compression: zstd` takes the conventional fast path and desyncs identically, so resolution belongs at the parse boundary.

## Known-open follow-up

An upstream applying zstd at the **transport** level (`Content-Encoding: zstd` over an uncompressed body — the shape NixOS/nix#10275 describes) still leaves ncps holding a raw NAR, because `upstream.GetNar` unconditionally sends `Accept-Encoding: zstd` and transparently strips whatever comes back. This PR makes the labels agree but cannot make those bytes true.

That case is covered by a **deliberately skipped subtest** so it stays visible rather than forgotten; the follow-up (not requesting transport zstd for content the narinfo already declares compressed) should un-skip it with no other edit. It is unclear whether the reporter is on that variant — awaiting confirmation on the issue.

## Test plan

- [x] `task fmt`, `task lint` (0 issues), `task test` — all green
- [x] `TestAtticNarInfoURLWithoutCompressionExtension` — Attic-shaped opaque URL: rows agree, URL is `nar/<narhash>.nar.zst`, served bytes are a single zstd frame warm and cold (was double-compressed)
- [x] `TestBareNarURLWithConventionalHashResolvesCompression` — same defect on the non-opaque path, incl. asserting the extension-less upstream path is persisted to `narinfos.upstream_url`
- [x] `pkg/nar` table tests — precedence, `none`/absent/unrecognised declarations, snix-castore `.nar`-less shape, and a no-panic guard
- [x] Existing cachix / snix-castore / nix-serve / cache.nixos.org tests unchanged

Refs #1470

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,6 +135,21 @@ project loosely follows [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- **Upstream narinfos that declare their compression only in the `Compression:`
+  header are no longer desynced from the NAR ncps stores.** ncps derived a NAR's
+  compression exclusively from the file extension on the narinfo `URL:` field.
+  Attic states compression only in the header — its URL is always a bare
+  `nar/<storePathHash>.nar` — so ncps read `none`, linked a `Compression: zstd`
+  narinfo to a `compression=none` `nar_file` row, re-compressed the
+  already-compressed NAR a second time on disk, and advertised a URL whose bytes
+  did not match the `Compression:` it re-signed. The narinfo header is now
+  authoritative whenever the URL carries no compression extension (an explicit
+  extension still wins, so cache.nixos.org, cachix, Harmonia and nix-serve are
+  unaffected), and the original upstream path is preserved for the upstream `GET`
+  so the NAR is still fetched from where the upstream actually serves it. Note
+  that an upstream applying zstd at the *transport* level (`Content-Encoding`)
+  over an uncompressed body is a separate, still-open case. (#1470)
+
 - **snix-castore (and other `.nar`-less opaque) upstream narinfo URLs are now
   proxied instead of returning `HTTP 500 "invalid nar URL"`.** Upstreams such as
   `cache.snix.dev` serve narinfos whose `URL:` field is a content-addressed
--- a/openspec/changes/archive/2026-08-18-narinfo-compression-authoritative/.openspec.yaml
+++ b/openspec/changes/archive/2026-08-18-narinfo-compression-authoritative/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-08-18
--- a/openspec/changes/archive/2026-08-18-narinfo-compression-authoritative/design.md
+++ b/openspec/changes/archive/2026-08-18-narinfo-compression-authoritative/design.md
@@ -1,0 +1,132 @@
+## Context
+
+See proposal.md — Why. The mechanics that shape the approach:
+
+`nar.ParseUpstreamURL(u, fallbackHash)` (`pkg/nar/url.go:64`) derives compression
+purely from the URL's file extension via `parseURLParts`, which has no access to
+the narinfo. It has two callers, both of which already hold the parsed narinfo:
+`pullNarInfo` (`cache.go:4276`) and `lookupPreferredUpstreamURL` (`cache.go:1821`).
+
+Downstream, three places consume that compression and are all currently fed the
+wrong value for a bare-`.nar` URL:
+
+- `pullNarInfo`'s normalization switch — the `case narURL.IsOpaque()` arm
+  (`cache.go:4368`) rewrites the served URL as
+  `nar.URL{Hash: narURL.Hash, Compression: narURL.Compression}` under a comment
+  claiming it is "preserving compression".
+- `storeInDatabase` (`cache.go:5641`) re-parses the *rewritten* `narInfo.URL` with
+  the strict `nar.ParseURL` to key the `nar_file` row.
+- `putNarInStore` (`cache.go:2237`) re-compresses the body as zstd when it believes
+  the compression is `none`.
+
+Because `storeInDatabase` re-derives from the emitted URL rather than from the
+resolved value, the emitted URL and the resolved compression must be kept in
+agreement — that constraint drives Decision 2.
+
+Attic's URL uses the 32-char store path hash, which fails `ValidateHash`, so it
+takes the opaque branch. A conforming upstream could equally serve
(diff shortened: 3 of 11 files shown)


kalbasit:

- #1474 ◀

kalbasit:
Addressed all three threads in c61af593.

**Query dropped for hash-named URLs** (`cache.go#L4373-L4383` / `#L4325-L4329`) — confirmed, and a genuine regression from this PR. Marking a conventional hash-named URL opaque (so its extension-less upstream path survives for the upstream GET) routed it through the rewrite branch that drops the query. That branch is correct for *truly* opaque URLs, whose query belongs only to the upstream GET, but a hash-named URL owns its query and `nar_file` rows are keyed by `(hash, compression, query)`. Fixed at both sites by distinguishing the two cases via ncps's own strict hash-named parser, with `TestBareNarURLWithQueryPreservesQuery` asserting the query survives onto the re-served URL, the `nar_file` row, and the upstream GET.

I rejected the obvious implementation of adding a field to `nar.URL` — it is unexported, so it changes struct equality against every `nar.URL` literal in the existing tests, for no behavioural gain.

**Propagate the resolved compression into `narInfo.Compression`** — real, but pre-existing and deliberately out of scope, so no behaviour change:

- It is not introduced here. For `URL: nar/<hash>.nar.xz` with `Compression: zstd`, the URL extension wins both before and after this PR (`resolveCompression` is a no-op when the URL carries an extension), so the resulting narinfo is byte-identical to `main`'s.
- It only arises for a **self-contradictory upstream** — one whose URL extension and `Compression:` header disagree. No known upstream does this.
- The suggested fix carries real risk. Assigning `narInfo.Compression = narURL.Compression` before the switch would stop `.nar.xz` + `Compression: none` from taking the normalization branch, so ncps would advertise `Compression: xz` while retaining `FileHash`/`FileSize` — which for a none-declared narinfo equal `NarHash`/`NarSize`. That means advertising a compressed NAR with the *uncompressed* size, breaking clients that check it. Today that case normalizes to `none` and is served correctly by the transparent-decompression fallback.

What I did instead: the requirement wording over-promised, so it now states explicitly that reconciling a contradictory URL/header pair is out of scope and that an unrecognised declared compression is ignored rather than trusted. Behaviour is unchanged; the spec now matches it.

**DevSkim "TODO or similar left in source code"** — false positive. The flagged `FIXME` was inside a doc comment quoting attic's own source (`file_hash: None, // FIXME`), not a marker on ncps code. Reworded so the trigger is gone at source rather than suppressed.

`task fmt`, `task lint` (0 issues) and `task test` are green.

kalbasit:
Verified, and it is accurate — but it is pre-existing, and the proposed fix is larger than it looks. Addressed as a spec-accuracy correction in 27469850 rather than a code change.

**The finding is correct.** `OpaqueUpstreamRef()` builds the query with `u.Query.Encode()` (`pkg/nar/url.go#L285`), which sorts keys and canonicalises percent-escapes, so `?b=1&a=%2f` is refetched as `?a=%2F&b=1`.

**Two corrections to the framing, though:**

1. **Fixing `OpaqueUpstreamRef` alone would not fix the refetch.** `JoinURL` — which builds the actual upstream GET — does the same `u.Query.Encode()` at `pkg/nar/url.go#L323`. Both would need a raw-query field threaded through them, plus a persistence round-trip through `upstream_url`/`WithOpaqueUpstreamRef`. That matches your own "heavy lift" label and is why I have not done it here.

2. **This predates this PR.** Both call sites arrived with the snix-castore opaque-URL support in 2880f19e (#1427), as did the word "verbatim" in the two pre-existing scenarios at `spec.md#L60` and `#L84`. This PR widens *which* URLs flow through that code; it does not introduce the normalisation.

**Real-world impact for the upstreams actually involved is nil:** attic's URL carries no query at all, and snix-castore's `?narsize=N` is invariant under canonical re-encoding (single key, digits only).

I also considered adding a raw-query field to `nar.URL` and rejected it for the same reason I rejected one in the previous round: it is unexported, so it changes struct equality against every `nar.URL` literal in the tests — including the existing `TestParseURL` cases that carry a `?hash=…` query.

So the thing that was actually wrong here is the **documentation**, and I have fixed that: the requirement now states that path preservation is byte-exact while a query is preserved *semantically* (parsed and re-encoded canonically, so key order and escape casing may differ), and that an upstream requiring a byte-exact query — e.g. one carrying a signature over the raw string — is unsupported.

If you want the byte-exact behaviour, it is worth its own change against `pkg/nar` covering `OpaqueUpstreamRef`, `JoinURL` and the persistence round-trip together, and it should also correct the pre-existing "verbatim" wording at `spec.md#L60`/`#L84`. Happy to open that separately.

PR #1478, opened 2026-08-27
fix(migrate): emit progress line before first tick

kalbasit:
## Summary

The chunks-to-nar progress reporter was driven purely by a ticker, so a migration
that finished inside a single interval emitted no progress line at all — the ticker
never fired.

That is a real reporting gap for short migrations, and it made
`TestMigrateChunksToNar_CLI_ProgressLogEmitted` non-deterministic. With the interval
overridden to 1ms and a one-item migration, the work raced the first tick and lost
roughly **four runs in five**. Verified on a clean tree with unrelated changes
stashed, so this is pre-existing on `main` and not caused by anything stacked above it.

## Approach

Extract the emission into a `logProgress` closure and call it once up front when there
is work to do, then once per tick as before.

- Guarantees at least one progress line whenever `total > 0`, which removes the race.
- Gives operators immediate feedback instead of a silent first interval.
- Gated on `total > 0` so an empty run stays completely silent, preserving
  `TestMigrateChunksToNar_CLI_NoProgressLogOnEmptyRun`.

## Test plan

- [x] `task fmt` exits 0
- [x] `task lint` exits 0
- [x] `task test` exits 0
- [x] 12/12 consecutive runs of both progress tests pass (was 4/5 failing before)

### DIFF
--- a/pkg/ncps/migrate_chunks_to_nar.go
+++ b/pkg/ncps/migrate_chunks_to_nar.go
@@ -327,39 +327,58 @@ func migrateChunksToNarAction(registerShutdown registerShutdownFn) cli.ActionFun
 
 		progressWg.Add(1)
 
+		// logProgress emits one progress line from the counters as they stand.
+		logProgress := func() {
+			elapsed := time.Since(startTime)
+			processed := atomic.LoadInt32(&totalProcessed)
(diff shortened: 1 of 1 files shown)


kalbasit:

- #1478 ◀
    - #1479

PR #1479, opened 2026-08-27
fix(cache): bound the NAR storage presence probe

kalbasit:
## Summary

Production ncps answered NAR requests with `HTTP 200` and a truncated body. Clients
reported `Truncated zstd input` and `curl error 92: HTTP/2 stream reset by server
(INTERNAL_ERROR)`.

Root cause, from a goroutine dump captured mid-stall against v0.10.0-rc17:

```
goroutine 366 [syscall]:
syscall.Syscall6(0x106, ...)                    <- fstatat
os.Stat(...)
  local.(*Store).StatNar        local/local.go:367
  cache.(*Cache).statNarInStore cache/cache.go:4954
  cache.(*Cache).GetNar         cache/cache.go:1294
```

A single `os.Stat` on the NFS mount blocked **~57s**, with the pod otherwise idle (81
goroutines, exactly one in `[syscall]`). The ingress read timeout is 60s, so nginx
aborted the response mid-body. A **2,021-byte** NAR took 56.87s to first byte, so this
is not size or bandwidth.

`os.Stat` bottoms out in `fstatat(2)`, which takes no context and cannot be aborted
from userspace, so a deadline cannot *cancel* the local probe — only stop the request
from *waiting* on it.

## Approach

- Bound the probe at the cache layer, not per backend: run it on its own goroutine and
  select over result, deadline and caller cancellation.
- Propagate the deadline into the backend context, so S3 (whose `StatObject` honours
  context) genuinely cancels while local is merely abandoned.
- A timed-out probe yields `ErrStatTimeout` — **undetermined, not a confirmed absence**.
  Enforced on both exit paths: upload-only mode must not return `storage.ErrNotFound`
  (that would tell `nix copy` to skip the upload and leave a phantom NAR), and the
  ordinary read path must not either — upstream recovery failing does not establish
  that the local copy is absent when the local probe never answered.
- Collapse concurrent probes for the same object with `singleflight` (20 callers → 1
  backend probe) and cap total in-flight probes, so a storage brown-out degrades
  instead of pinning an OS thread per client.
- Add `cache.storage.stat-timeout` (default 5s, `0` disables for rollback).
- Export `ncps_storage_stat_duration_seconds`, `_timeout_total`, `_in_flight`.

## Why this survived so long

Every previous fix targeted *which bytes get served*. None bounded *how long a waiter
may sit silent*. The existing staging-contention scenario reads NARs with a **900s**
client timeout and asserts only that bytes match — so a 57s stall scored as a PASS.

This PR adds time-to-first-byte measurement to the e2e harness and asserts it against a
budget on every warm NAR read.

## Measured effect

| | result |
|---|---|
| Before | `GetNar` never returned; test failed after exhausting its 10s budget |
| After | `GetNar` resolved in **1.00s** against a 30s uncancellable probe |

Full evidence, including the goroutine dump, is in
`openspec/changes/archive/2026-08-27-fix-nar-serve-stall-proxy-timeout/`.

## What this does NOT fix

The substrate. `ncps_storage_stat_timeout_total` is the signal: non-zero in production
means storage is still stalling and this is converting stalls into upstream fallbacks
rather than truncated responses. NFS mount tuning and the move to the S3 backend are
tracked as infrastructure work.

## Test plan

- [x] `task fmt` exits 0
- [x] `task lint` exits 0
- [x] `task test` exits 0
- [x] `nix build .#checks.x86_64-linux.e2e-harness-unit` — 70 passed
- [x] `nix run .#e2e -- --mode local --scenario single-local-sqlite` — PASS, `warm NAR ttfb=0.001s (budget 15.0s)`
- [x] `openspec validate --specs --strict` — 47 passed, 0 failed

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -133,6 +133,17 @@ cache:
   trusted-upload-keys: []
   #   - my-cache-1:abcdef0123456789...=
   storage:
+    # Maximum time the NAR read path waits on a storage presence probe before
+    # treating presence as undetermined (default: 5s). Keep it well below your
+    # reverse-proxy read timeout. The local backend's probe is an os.Stat, which
+    # bottoms out in an uncancellable fstatat(2): on a hard NFS mount a single one
+    # has been measured blocking ~57s, long enough for the proxy to abort the
+    # response mid-body and hand the client a 200 with a truncated body. 5s sits an
+    # order of magnitude above a healthy probe (8-300ms) and an order of magnitude
+    # below a 60s proxy timeout, so it fires only on genuine pathology.
+    # A timed-out probe is treated as undetermined, never as a cache miss, so it
+    # never becomes a spurious 404. Set to 0 to disable the bound (rollback).
+    stat-timeout: 5s
     # The local data path used for configuration and cache storage
     # Use this OR S3 storage (cache.storage.s3.bucket) - not both
     local: "/var/lib/ncps"
--- a/nix/e2e-tests/src/client.py
+++ b/nix/e2e-tests/src/client.py
@@ -14,11 +14,24 @@ import lzma
 import os
 import subprocess
 import sys
+import time
 import urllib.error
 import urllib.request
+from dataclasses import dataclass
 from typing import Dict, Optional, Tuple
 
 from harness_config import REPO_ROOT, VAR_NCPS
+
+
+@dataclass
+class TimedResponse:
+    """An HTTP response with its time-to-first-byte recorded."""
+
+    status: int
+    headers: Dict[str, str]
+    body: bytes
+    ttfb_seconds: float
+    total_seconds: float
 
 
 class Client:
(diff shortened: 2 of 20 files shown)


kalbasit:

- #1478
    - #1479 ◀

kalbasit:
Addressed the merge-risk findings. Verified each one against the code before changing anything rather than applying them on faith — one was real and material, one I'm pushing back on, one is already bounded.

### 1. "spend multiple probe timeouts within one request" — CONFIRMED, fixed

This was right, and worse than it sounds. I measured it: bounding each *probe* is not the same as bounding the *request*, because a single `GetNar` consults the store several times (the pre-check, the servability lookup, and again after download coordination).

| | `GetNar` elapsed against a 300 ms bound |
|---|---|
| before | **1.20 s — 4.0x** |
| after | **0.30 s — 1.0x** |

At the 5 s default that was 20 s, not 5 s. At a 15 s setting it would have pushed a request back over a 60 s proxy read timeout — reintroducing the exact production failure this PR exists to fix.

Fixed with a cumulative per-request probe budget carried on the context: every probe spends from one deadline, and an exhausted budget returns `ErrStatTimeout` immediately instead of starting another wait. Deliberately scoped to probes only — it must not bound the download, which legitimately takes far longer than any probe should.

Pinned by `TestRequestProbeBudgetIsCumulative`, which fails if the total ever scales with the number of probes.

### 2. "pre-launch concurrency cap" — pushing back

The cap is intentionally checked *inside* the singleflight function rather than before `DoChan`, and I think moving it earlier would be a regression.

`DoChan` runs the function **once per key**. Callers that arrive while a probe for the same key is already in flight do not start a probe — they join the existing one. Checking the cap before `DoChan` would reject those joiners even though they cost no additional goroutine and no additional blocked syscall, which is precisely the case the cap should permit. Checking inside means the counter tracks *distinct in-flight probes*, which is the resource actually being bounded.

A goroutine is created for a rejected key, but it returns immediately without touching the backend, so it does not accumulate. Happy to reconsider if there's a failure mode I'm not seeing.

### 3. "uncapped upstream recovery after a timeout" — already bounded

Upstream recovery is not unbounded. `setupHTTPClient` sets `ResponseHeaderTimeout` (`upstream.response-header-timeout`, 3 s in the deployment that hit this) and a dial timeout on the transport, so time-to-first-byte from an upstream is capped per attempt. Adding a second deadline on top would duplicate an existing, separately tunable control.

I did not expand scope to the retry/multi-upstream accumulation, since that is pre-existing behaviour of the upstream path rather than something this PR introduces.

---

`task fmt` / `task lint` / `task test` all exit 0; `openspec validate --specs --strict` 47/47. The spec required a request-level bound all along ("a NAR request SHALL either begin emitting response body bytes, or terminate with an explicit error status, within a configured budget") — the contract was right and the implementation was not, so the spec is unchanged apart from an added scenario making the cumulative case explicit.

PR #1492, opened 2026-09-06
fix(upstream): negotiate transport zstd only for uncompressed NARs

kalbasit:
## Summary

Closes the last open gap from #1470 — the one the quarantined subtest described.

ncps sent `Accept-Encoding: zstd` on **every** upstream NAR fetch and transparently stripped any `Content-Encoding: zstd` it got back. For a NAR the narinfo already declares compressed, that request buys effectively nothing (zstd over zstd/xz yields ~0%) while inviting a compressing proxy in front of the upstream — Caddy's `encode zstd`, nginx, a CDN — to wrap the body. Where the upstream's own content was **not** in fact compressed, ncps was left holding a raw NAR that it stored and served as `Compression: zstd`, and nix failed with `input compression not recognized`.

What made ncps uniquely exposed: it always advertised zstd, while nix's own curl frequently is not built with it. So such a proxy compressed **for ncps alone**, and the same store path substituted fine directly — the exact asymmetry reported in #1470.

## What changed

Transport zstd is now negotiated only for a NAR whose resolved compression is `none` — the Harmonia shape the negotiation exists for. The gate lives in `GetNar` beside the matching decompression, so the two read together instead of being scattered across packages.

**Transparent decompression stays unconditional.** A transfer encoding is a property of the response, not of what we asked for; refusing to decode one would turn a working fetch from an eager-but-conforming upstream into a corrupt one. This change removes the case where ncps *invited* the encoding — it does not police upstreams that send it unbidden. `TestGetNarDecompressesUnsolicitedContentEncoding` pins that, so a future "only strip what we negotiated" simplification fails loudly.

**Dependency worth noting:** this only works because of the archived `narinfo-compression-authoritative` change. Before it, an Attic-shaped narinfo resolved to `none` despite zstd content — so gating on `Compression == none` would have mis-fired on precisely the shape that motivates this.

## The quarantined subtest is un-skipped

Its fake upstream set `Content-Encoding: zstd` **unconditionally**, modelling an upstream that lies regardless of the request. No real proxy does that — Caddy, nginx and CDNs all compress only when the client advertises — and left that way the subtest could not pass under *any* correct implementation, because it never reacted to what ncps asked for. Making the fixture realistic is what lets it run.

That does narrow what it proves: an upstream sending an **unsolicited** `Content-Encoding: zstd` over a raw body is a protocol violation that remains unsupported by design, and is no longer represented by a test. Called out in `design.md` — Risks.

`pkg/cache/attic_narinfo_url_compression_test.go` now has zero skips.

## Test plan

- [x] `task fmt`, `task lint` (0 issues), `task test` — green, before and after restacking onto current `main`
- [x] `TestGetNarAcceptEncodingMatchesContentCompression` — `none` negotiates; `zstd`/`xz` do not (the `none` row passed before the fix, so the test is not vacuous)
- [x] `TestGetNarDecompressesUnsolicitedContentEncoding` — unconditional strip preserved
- [x] `TestAtticNarInfoURLWithoutCompressionExtension` — both subtests pass, no skips
- [x] Harmonia (`Compression: none` + transport zstd), nix-serve, cachix and cache.nixos.org paths unchanged; `pkg/server` downstream `Accept-Encoding` untouched

Refs #1470

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,6 +135,21 @@ project loosely follows [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
+- **A compressing reverse proxy in front of an upstream no longer corrupts
+  already-compressed NARs.** ncps sent `Accept-Encoding: zstd` on *every*
+  upstream NAR fetch and transparently stripped any `Content-Encoding: zstd` it
+  got back. For a NAR the narinfo already declared compressed that bought
+  effectively nothing — zstd over zstd/xz yields ~0% — while inviting a proxy
+  (Caddy's `encode zstd`, nginx, a CDN) to wrap the body. Where the upstream's
+  own content was not in fact compressed, ncps was left holding a raw NAR that it
+  stored and served as `Compression: zstd`, and nix failed with `input compression not recognized`. ncps was uniquely exposed because it always
+  advertised zstd while nix's own curl frequently is not built with it, so such a
+  proxy compressed for ncps alone and the same store path substituted fine
+  directly. Transport zstd is now negotiated only for NARs whose content is
+  uncompressed (the Harmonia shape the negotiation exists for); a
+  `Content-Encoding: zstd` response is still transparently decompressed whether
+  or not it was solicited. (#1470)
+
 - **Upstream narinfos that declare their compression only in the `Compression:`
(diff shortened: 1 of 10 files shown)


kalbasit:

- #1491
    - #1492 ◀

